PR #126, opened 2016-09-02
Update docker-utils based on review comments

arrawatia:
Changes based on review comments from @hachikuji and @ijuma .

@theduderog @cotedm Can you please review this PR ?

Review comments : https://github.com/confluentinc/docker-utils/pull/4

### DIFF
--- a/java/src/main/java/io/confluent/admin/utils/cli/KafkaReadyCommand.java
+++ b/java/src/main/java/io/confluent/admin/utils/cli/KafkaReadyCommand.java
@@ -102,9 +102,9 @@ public class KafkaReadyCommand {
                 if (res.getString("config") != null) {
                     workerProps = Utils.propsToStringMap(Utils.loadProps(res.getString("config")));
                 }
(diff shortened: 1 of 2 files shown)


sophieqianCA:
@arrawatia  I saw you contributed to https://github.com/confluentinc/cp-docker-images/blob/master/java/src/main/java/io/confluent/admin/utils/ClusterStatus.java.  

Right now in our QA cluster environment (my dev cluster is working with no issue), TLS is enabled for Kafka, we are getting  following error when starting SR, "Expected 1 brokers but found only 0. Trying to query Kafka for metadata again ..." . I was able to verify the cluster works correctly under non TLS mode,  and verified every single node ( without put them in cluster) is working correctly under TLS mode. 

Can you provide me some directions on what I should look at to further debug this issue? Why AdminCLient can not find any brokers?   Thanks   -- sophie

sophieqianCA:
OK, I got this issue resolved, our QA server hostname is not using FQDN, I asked them to change to use FQDN name, everything works now.

PR #132, opened 2016-09-14
First version of new example: single-node complete CP deployment

dbtucker:
Now that docker-for-mac is more stable, this example allows a standalone deployment of 
ALL the platform components (including a Control Center that will enable Connector config).

Happy to add details to the doc pages if necessary.


coughman:
I have a working rest-proxy compose config.  I'll pile my changes on once this is merged

PR #134, opened 2016-09-15
docker compose example for CP

coughman:
has schema registry and rest proxy


samjhecht:
@coughman - @dbtucker opened a PR yesterday for a full-CP docker compose example: https://github.com/confluentinc/cp-docker-images/pull/132

It looks like his doesn't have Rest Proxy but it does have connect and .  You want to just add to that one instead?  Or vice versa?

coughman:
sure I will pile on my rest proxy changes on top of @dbtucker's changes.  Closing this PR then

PR #136, opened 2016-09-19
3.0.1 docs

samjhecht:


samjhecht:
creating a new branch for 3.0.1 docs.  from now on, anything pushed to master will need to be moved here in order to end up on the CP docs.

PR #139, opened 2016-09-20
Kafka rest configs

samjhecht:


samjhecht:
this will be fixed in another PR.

PR #141, opened 2016-09-24
Fix double log4j logging

dsmiley:
The root logger (line 2) already has "stdout", and thus it's inherited automatically by sub-loggers.  By specifying them in sub-loggers it results in twice the log messages.


ewencp:
Hi @dsmiley, would you mind signing our [CLA](http://clabot.confluent.io/cla)? We use it to make sure licenses are handled properly for contributions so they can be redistributed under [Apache 2.0](https://github.com/confluentinc/cp-docker-images/blob/master/LICENSE).

dsmiley:
I signed the CLA just now.

samjhecht:
Thanks @dsmiley .  I'm merging this PR now.

PR #155, opened 2016-10-24
Updates to c3 image

theduderog:
@samjhecht @jonchiu 

- new config params
- renamed to enterprise-control-center
- updated smoke tests to match new APIs


jonchiu:
🚀 🔦

PR #159, opened 2016-11-02
Bump versions and fix some dir names

jonchiu:


jonchiu:
@theduderog @samjhecht

samjhecht:
good catch, @jonchiu .  LGTM

PR #160, opened 2016-11-03
Updates to support running C3 against a secured cluster

jonchiu:
Passes a property files with settings for SSL support to `cub kafka-ready` which enables the C3 container to boot properly.


jonchiu:
@theduderog

theduderog:
Let's add a test for admin.properties and then LGTM.  Thanks, @jonchiu

PR #161, opened 2016-11-03
Fix issue with Mesos deployment

dbtucker:
Need proper setting of KAFKA_REST_HOST when in mesos environments (same logic 
as done with other images).

### DIFF
--- a/debian/kafka-rest/include/etc/confluent/docker/apply-mesos-overrides
+++ b/debian/kafka-rest/include/etc/confluent/docker/apply-mesos-overrides
@@ -19,6 +19,6 @@
 # 
 # Use those values provide things we know we'll need.
 
-[ -n "${HOST:-}" ] && [ -z "${KAFKA_REST_HOST:-}" ] && \
-	export KAFKA_REST_HOST=$HOST || true # we don't want the setup to fail if not on Mesos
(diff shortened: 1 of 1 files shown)


debian/kafka-rest/include/etc/confluent/docker/apply-mesos-overrides
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Mesos DC/OS docker deployments will have HOST and PORT0 
+# set for the proxying of the service.
+# 
+# Use those values provide things we know we'll need.
+
+[ -n "${HOST:-}" ] && [ -z "${KAFKA_REST_HOST:-}" ] && \
+	export KAFKA_REST_HOST=$HOST || true # we don't want the setup to fail if not on Mesos
```
samjhecht: KAFKA_REST_HOST_NAME?

dbtucker: You're right ... that's what I get from testing the fix against the wrong branch.

samjhecht:
LGTM otherwise @dbtucker

PR #165, opened 2016-11-08
Make SR and RP config consistent.

### LINKED ISSUE #163: Make SR and RP config template same as Kafka, Connect and C3
SR and CP has a fix list of params in the config template[1]. To make it consistent with other images we should use the wildcard function[2].

The zookeeper image will still have a fixed list because the ZK naming scheme for parameters is inconsistent[3].

1. https://github.com/confluentinc/cp-docker-images/blob/master/debian/kafka-rest/include/etc/confluent/docker/kafka-rest.properties.template
2. https://github.com/confluentinc/cp-docker-images/blob/master/debian/kafka-connect-base/include/etc/confluent/docker/kafka-connect.properties.template
3. https://github.com/confluentinc/cp-docker-images/blob/master/debian/zookeeper/include/etc/confluent/docker/zookeeper.properties.template#L16-L18

cc @dbtucker @cjmatta

arrawatia:
Fixes #163


samjhecht:
LGTM

PR #168, opened 2016-11-10
JMX Warning message refers to wrong env var

dsmiley:
The warning message refers to KAFKA_OPTS but as you can see from the condition, the env var KAFKA_JMX_OPTS is what's pertinent.


ewencp:
ok to test

arrawatia:
@dsmiley Thanks for the PR. Merging now.

Issue #176, opened 2016-11-22
Kafka-ready blocks security enablement for Schema Registry Image


samjhecht:
Security support was added to SR in 3.0.1, but SR's cub kafka-ready check currently restricts enablement of security for SR w/ secure kafka. 

kafka-ready currently supports 2 options:
1. get broker list from ZK (for legacy reasons related to SR & RP)
2. you give the broker list and security config

Given that Schema Registry now supports secure clusters and has a bootstrap.list, we'll need to switch to using the second option.  

we can do the following: 
- remove this blocker for SR much as we do for Control Center by replacing: https://github.com/confluentinc/cp-docker-images/blob/master/debian/schema-registry/include/etc/confluent/docker/ensure#L26
- with: https://github.com/confluentinc/cp-docker-images/blob/master/debian/enterprise-control-center/include/etc/confluent/docker/ensure#L27-L30

We'll need to add the following:
```
cub kafka-ready 1 \
  "${SCHEMA_REGISTRY_CUB_KAFKA_TIMEOUT:-40}" \
  -b "${SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS}" \
  --config /etc/"${COMPONENT}"/admin.properties
```

and we'll need to add an admin properties file for SR like this one: https://github.com/confluentinc/cp-docker-images/blob/master/debian/enterprise-control-center/include/etc/confluent/docker/admin.properties.template

```
{% set security_props = env_to_props('SCHEMA_REGISTRY_KAFKASTORE', '') -%}
{% for name, value in security_props.iteritems() -%}
{{name}}={{value}}
{% endfor -%}
```

Then we need to add a like like this @ the configure script: https://github.com/confluentinc/cp-docker-images/blob/master/debian/enterprise-control-center/include/etc/confluent/docker/configure#L36

Finally, we need tests!

sophieqianCA:
Hi, @samjhecht :  I build the docker image with the change suggested by you,  got following error:
admin.properties file not been generated from template.

echo "===> Check if Kafka is healthy ..."
===> Check if Kafka is healthy ...

cub kafka-ready 1 \
  "${SCHEMA_REGISTRY_CUB_KAFKA_TIMEOUT:-40}" \
  -b "${SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS}" \
  --config /etc/"${COMPONENT}"/admin.properties
[...]

Issue #185, opened 2016-12-13
Add unit tests for secure SR , Connect and C3


arrawatia:
Add some tests like these for all components that support security : 

- https://github.com/confluentinc/cp-docker-images/blob/master/tests/test_kafka.py#L270
- https://github.com/confluentinc/cp-docker-images/blob/master/tests/test_kafka.py#L406
- https://github.com/confluentinc/cp-docker-images/blob/master/tests/test_kafka.py#L460

PR #193, opened 2016-12-20
kafka-rest: Fix for NullReferenceException in kafka-ready (SSL enabled)

amccague:
This is a change similar to that for issue #176 but for Kafka Rest. Following a similar approach.

This allows me to get past the `kafka-ready` step and use an `http` or `https` listener with an SSL Kafka cluster.


samjhecht:
@amccague thanks for the PR!  This looks good.  However, we're going to need to have tests for this (as well as for secure SR and Connect) before we will upload new images.  If you want to take a stab at writing them, that would be most welcome.  The scaffolding is all there and we have the example of the Kafka tests: https://github.com/confluentinc/cp-docker-images/issues/185.

Otherwise, we're planning to add the tests on our end before long.

arrawatia:
Merging this for now as we need the fix. 

Please feel free to open a new PR with tests :).

Issue #190, opened 2016-12-15
Docs: Insufficient details on security configuration


dbtucker:
The security.rst file mentions that the secrets for the deployment can be loaded via a Docker mounted volume ... but there are no more specifics.   This needs a step-by-step discussion of what files are needed and how they are used.

Additionally, there is no example of deploying the images with arbitrary user-id's.

samjhecht:
Good catch.  We need to add links to our docker-specific tutorials in this document: http://docs.confluent.io/3.1.1/cp-docker-images/docs/security/security.html.  

In the meantime, you can find this information in these tutorials:
* http://docs.confluent.io/3.1.1/cp-docker-images/docs/tutorials/clustered-deployment-sasl.html
* http://docs.confluent.io/3.1.1/cp-docker-images/docs/tutorials/clustered-deployment-ssl.html

dbtucker:
Thanks, Sam.   The tutorials do cover the info I was looking for.

Issue #180, opened 2016-12-08
Make ZOOKEEPER_CLIENT_PORT default to 2181.


arrawatia:
This is recommended default value in the Confluent distro.

PR #200, opened 2017-01-13
Issue 180, 190 and docs improvements

samjhecht:
* Fix issue reported in #180 - make ZOOKEEPER_CLIENT_PORT default to 2181

* fix links to security tutorials per https://github.com/confluentinc/cp-docker-images/issues/190

* remove specific AMI from development docs `docker-machine create` example.  The one we included was no longer available and it's not necessary to specify it anyway.


arrawatia:
LGTM.

Issue #49, opened 2016-07-23
Test on docker overlay network


arrawatia:
Add tests for overlay network in addition to tests for bridged and host networks. 

This will involve setting up a swarm cluster with overlay networks. We can modify the existing multinode tests to include swarm specific properties to ensure that the containers are deployed on multiple nodes.

Add tests to find out the impact of overlay networks on performance.

References: 
- https://docs.docker.com/engine/userguide/networking/get-started-overlay/
- https://docs.docker.com/compose/swarm/
- https://docs.docker.com/compose/networking/

PR #201, opened 2017-01-16
Create example of Confluent stack with overlay networks

afolarin:
This adds an example deployment of Confluent components on Swarm, using overlay networks. Might be a starter to address confluentinc/cp-docker-images#49

**Caveats** 
- docker 1.13.x and above (for the `docker deploy --compose-file` option)
- can't create attachable networks in the docker-compose.yml file
- no idea about performance (but probably not great)
- there seem to be issues with multiple zookeeper instances RADAR-CNS/RADAR-Docker#7

### DIFF
--- a/examples/cp-kafka-overlay/README.md
+++ b/examples/cp-kafka-overlay/README.md
@@ -1,0 +1,37 @@
+# RADAR-CNS with multi-node cluster using Docker Swarm 
+
+
+# If you require attachable networks for testing, you'll need to create them as external e.g.
+```sh
+docker network create --attachable --driver overlay zookeeper
+docker network create --attachable --driver overlay kafka
+docker network create --attachable --driver overlay api
+```
+
+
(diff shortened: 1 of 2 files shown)


samjhecht:
Hi @afolarin - thanks for submitting this PR.  For examples and tutorials, we typically try to have more documentation on what everything is and all of the set up steps.  We also add verification steps along the way after each step so the user knows whether they did things right or not.  Here's an example of one of the existing ones: http://docs.confluent.io/3.1.2/cp-docker-images/docs/tutorials/clustered-deployment.html

Do you think you could please expand on this example to follow that pattern?

afolarin:
@samjhecht yes sure. I'll follow your format from https://github.com/confluentinc/cp-docker-images/blob/f432909b3323ff033338c6f4bec36b555f8d3caa/docs/tutorials/clustered-deployment.rst

PR #204, opened 2017-01-20
Connect: Removed stdout from custom loggers so it doesn't output duplicate log

simplesteph:
this mimic the behaviour at
https://github.com/confluentinc/cp-docker-images/blob/master/debian/kafka/include/etc/confluent/docker/log4j.properties.template#L25


simplesteph:
@samjhecht while I got your attention can you please look at this PR ?

samjhecht:
This looks good.  Merging.

PR #209, opened 2017-01-25
Allow overrides in a local makefile

theduderog:
@arrawatia @samjhecht @jonchiu 

With this PR, you can put any custom build overrides you want in a local.make file.


jonchiu:
LGTM

PR #223, opened 2017-02-26
Add telnet

jomach:
I think telnet is  a good option to have on the containers. Specially for testing connections.


norwood:
@jomach can you give some insight to why you would want/need to have telnet on one of the images?

jomach:
Well to test connectivity between services is very Useful to just do docker exec -it .... bash and then try telnet

ewencp:
@jomach It looks like `nc` is already installed, would that not support your use case? `telnet` obviously isn't huge, but people seem to be quite sensitive to keeping images as thin as possible, thus the hesitation to add more stuff to the base image.

PR #224, opened 2017-02-28
Enhance cub to allow for alternate location for docker utility jar

dbtucker:
We can choose whether to document this or not.


norwood:
@dbtucker what is the use case for this? 

if its something good, it actually seems like a cleaner thing would be to export this env var via the Dockerfile and always use the env var in the *ub scripts

norwood:
ping

ewencp:
Closing since this is stale, in deprecated repo, and the use case was never explained. We can revive if someone finds a need for it.

PR #182, opened 2016-12-09
fix the other part of issue 176 

sophieqianCA:
make sure no unbound variable SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS in non TLS mode issue.

verified in both non TLS and TLS mode.

Please review, thanks

PR #233, opened 2017-03-23
Fix SSL issue for SR and RP

arrawatia:
- Apply changes from #182 and #193 
- Bump up the build no
- Better error message for null bootstrap brokers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-BUILD_NUMBER := 2
+BUILD_NUMBER := 3
 CP_VERSION := 3.2.0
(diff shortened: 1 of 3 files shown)


cotedm:
@arrawatia looks like this only address the schema registry, I think we need to also modify the REST proxy's ensure scripts in a similar way yes?

arrawatia:
@cotedm They are already on the 3.2.x branch. I merged #193 earlier.

cotedm:
Ahh makes sense @arrawatia thanks. LGTM pending test pass

PR #234, opened 2017-03-23
Update ensure for SR

cotedm:
@arrawatia @aayars @theduderog can you look at this? Looks like there's a typo in the ensure for SR that prevents startup with SSL enabled.


cotedm:
Fixed in #235

PR #240, opened 2017-03-29
C3 no longer depends on ZK, only used for ACLs script

rnpridgeon:
The only reason ZK connect exists within the configuration file as of 3.2 is to service the set ACL's script. That said there is no reason to make this configuration value mandatory.


norwood:
👍

ijuma:
Seems like this should be merged.

PR #244, opened 2017-04-05
dub: add `url-ready` command

miguno:


debian/base/include/dub
```diff
@@ -205,11 +252,15 @@ if __name__ == '__main__':
     check_env = actions.add_parser('ensure', description='Check if env var exists.')
     check_env.add_argument('name', help='Name of env var.')
 
-    check_env = actions.add_parser('wait', description='wait for network service to appear.')
+    check_env = actions.add_parser('wait', description='Wait for network service to appear.')
     check_env.add_argument('host', help='Host.')
-    check_env.add_argument('port', help='Host.', type=int)
+    check_env.add_argument('port', help='Port.', type=int)
```
miguno: Unrelated fix (wrong argument description)

debian/base/include/dub
```diff
@@ -205,11 +252,15 @@ if __name__ == '__main__':
     check_env = actions.add_parser('ensure', description='Check if env var exists.')
     check_env.add_argument('name', help='Name of env var.')
 
-    check_env = actions.add_parser('wait', description='wait for network service to appear.')
+    check_env = actions.add_parser('wait', description='Wait for network service to appear.')
```
miguno: Unrelated fix: Consistent formatting.

debian/base/include/dub
```diff
@@ -126,6 +128,51 @@ def wait_for_service(host, port, timeout):
...
+        port = 443
+
+    # Check if you can connect to the endpoint
+    status = wait_for_service(host, port, timeout)
+
+    if status:
+
+        # Check if service is responding as expected
+        r = requests.get(url)
+        if r.status_code // 100 == 2:
```
theduderog: is this any clearer than `r.status_code  == 200`?

debian/base/include/dub
```diff
@@ -205,11 +252,15 @@ if __name__ == '__main__':
...
     check_env.add_argument('name', help='Name of env var.')
 
-    check_env = actions.add_parser('wait', description='wait for network service to appear.')
+    check_env = actions.add_parser('wait', description='Wait for network service to appear.')
     check_env.add_argument('host', help='Host.')
-    check_env.add_argument('port', help='Host.', type=int)
+    check_env.add_argument('port', help='Port.', type=int)
     check_env.add_argument('timeout', help='timeout in secs.', type=float)
 
+    check_env = actions.add_parser('url-ready', description='Wait for a URL to be retrievable.')
```
theduderog: should we call this `http-ready`?  That's what it is.  We're not handling any other types of URLs

debian/base/include/dub
miguno: That part of the code is copied from `cub`'s `kr-ready`.  I actually considered to provide an optional parameter, `status_code` with a default of `200`, but eventually decided to keep things simple for the time being, and the idea to accept any `2xx` codes sounded like a good idea (which I assume was the reason for the related function in `cub`).

I'm happy to change this if need be.

debian/base/include/dub
miguno: +1

debian/base/include/dub
theduderog: oh, right.  makes sense as is.

theduderog:
@miguno Feel free to merge.  The jenkins build is badly configured and always fails :(

miguno:
Should I also merge this into `master`?  I have noticed that `3.2.x` hasn't been merged into `master` since 6 weeks, but I suppose it's fair game to just to a `git checkout master && git merge 3.2.x`, right?

theduderog:
yes, 3.2.x should be merged into master  cc @aayars

PR #247, opened 2017-04-06
kafka-connect-base: fix test condition for SSL

miguno:


miguno:
Subtle difference, but the original test condition for SSL did not check for `If the string equals "SSL"` -- it matched even if `CONNECT_SECURITY_PROTOCOL` was set to some other, non-empty string such as "foobar".

miguno:
Not sure whether any of the `tests/*` for kafka-connect-base would need updating, too.

arrawatia:
Wow ! That is subtle indeed. Unfortunately, we have no SSL/SASL tests for connect. Kafka/ZK are the only components with full coverage.

Shellcheck agrees. @theduderog is there a way to include shellcheck and pep-8 checks into the build process ?
image

theduderog:
@arrawatia Add shellcheck and pep8 steps to the Makefile?

PR #251, opened 2017-04-24
Update doc for running SASL cluster using docker-compose

parambirs:
The documentation for "Docker Compose: Setting Up a Three Node CP Cluster with SASL" was missing clear instructions for using docker-machine with docker-compose. It was also not clear how to build and run the kerberos image. I've attempted to clear up the confusion for these 2 points.


parambirs:
[clabot:check]

PR #253, opened 2017-04-25
Minor improvements

miguno:


docs/development.rst
```diff
@@ -380,76 +380,38 @@ The following examples show to extend the images.
 
   The images ship with Azul Zulu OpenJDK.  Due to licensing restrictions, we cannot bundle Oracle JDK, but we are testing on Zulu OpenJDK and do suggest it as a viable alternative.  In the event that you really need to use Oracle's version, you can follow the steps below to modify the images to include Oracle JDK instead of Zulu OpenJDK.
 
-  1. Change the base image to install Oracle JDK instead of Zulu OpenJDK.  First, you'll need to update the base image ``Dockerfile``:
+  1. Change the base image to install Oracle JDK instead of Zulu OpenJDK by updating ``debian/base/Dockerfile``.
```
miguno: Clarified the location because it wasn't sufficiently clear IMHO for the average reader.

docs/development.rst
```diff
@@ -380,76 +380,38 @@ The following examples show to extend the images.
...
-
-       # Confluent
-       ENV SCALA_VERSION="2.11"
-       ENV CONFLUENT_MAJOR_VERSION="3.3"
-       ENV CONFLUENT_VERSION="3.3.0-SNAPSHOT"
-       ENV CONFLUENT_DEB_VERSION="1"
-
-       # Zulu
-       ENV ZULU_OPENJDK_VERSION="8=8.15.0.1"
+       # In debian/base/Dockerfile
```
miguno: Essentially, what I changed was to show only the relevant snippet in `Dockerfile` that needs modification.  The problem with the previous text was that (a) it was a copy of the full, final `Dockerfile` which made it difficult to see which parts are actually relevant for the JDK change and (b) the "surrounding" code lines no longer match the latest `Dockerfile` in `master`.  Hence I prefer a diff-based approach that only covers the JDK related lines.

miguno: Alternatively, we could also provide a ready-to-use `Dockerfile` template for Oracle JDK, e.g. as `debian/base/Dockerfile.oracleJDK`?

PR #254, opened 2017-05-01
Use correct branch for Kafka Streams Examples

miguno:


miguno:
I created the `3.2.1-post` branch at [confluentinc/examples](https://github.com/confluentinc/examples) so that we can build the Docker images -- with the new changes in this PR -- in the correct way, without impacting users of the examples repository, whose default branch is `3.2.x` (and we can't bump that branch yet to use the not-yet-released 3.2.1 artifacts).

miguno:
Eventually, this PR should also go into `3.2.x` I think.

miguno:
Will you be merging this to 3.2.x eventually @aayars ?

aayars:
Sure, I can do that.

PR #256, opened 2017-05-02
lets try to dockerize centos :thumbsup:

norwood:
figured id just add the whole company on this.

basics here are creating a parallel Dockerfile structure for our services on centos. i also have an example in `cp-all-in-one` of overriding the compose script to use rpms (in fact its intermingling some rpm services and some debian services. what fun)

only really shitty thing here is that we have a bunch of copy/paste for the common stuff. its possible that with some smarter Makefile magic i could fix that.


norwood:
[clabot:check]

norwood:
ping

rnpridgeon:
pong?

aayars:
Cc: @maxzheng

Makefile
```diff
@@ -43,11 +47,23 @@ build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
...
 		fi; \
-		docker build --build-arg COMMIT_ID=${COMMIT_ID} --build-arg BUILD_NUMBER=${BUILD_NUMBER} $${BUILD_ARGS} -t confluentinc/cp-$${component}:latest debian/$${component} || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:latest  || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${CP_VERSION} || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${VERSION} || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${COMMIT_ID} || exit 1 ; \
+		for type in "" rpm; do \
+			DOCKER_FILE="debian/$${component}/Dockerfile"; \
+			COMPONENT_NAME=$${component}; \
+			echo $${type}; \
```
maxzheng: This echo nothing for "" ?

debian/base/Dockerfile.rpm
```diff
@@ -0,0 +1,94 @@
+#
```
maxzheng: It's a bit weird to have centos Dockerfile inside of the "debian" folder. 

Can we do either of the following? 
1. Create a new "centos" folder so we have centos/base/Dockerfile and then toggle everything based on "centos" vs "debian" (instead of "" vs "rpm"). Alternatively do "rpm/base" vs "deb/base". 
2. Or  reverse the grouping so we get "base/rpm/Dockerfile" vs "base/deb/Dockerfile"

Makefile
```diff
@@ -43,11 +47,23 @@ build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
...
+			if [ "$${type}" = "rpm" ]; then \
+				echo "this an rpm"; \
+				COMPONENT_NAME="rpm-$${component}"; \
+				DOCKER_FILE="$${DOCKER_FILE}.rpm"; \
+			fi; \
+			if [ -a "$${DOCKER_FILE}" ]; then \
+				docker build --build-arg COMMIT_ID=${COMMIT_ID} --build-arg BUILD_NUMBER=${BUILD_NUMBER} $${BUILD_ARGS} -t confluentinc/cp-$${COMPONENT_NAME}:latest -f $${DOCKER_FILE} debian/$${component} || exit 1 ; \
+				docker tag confluentinc/cp-$${COMPONENT_NAME}:latest ${REPOSITORY}/cp-$${COMPONENT_NAME}:latest  || exit 1 ; \
```
maxzheng: How come we use "confluentinc" in the source image while using $REPOSITORY in the target image even though they are the same value -- this is no op? Same with the build command and other tags below where 'confluentinc' is used instead of $REPOSITORY.

Makefile
```diff
@@ -43,11 +47,23 @@ build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
...
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:latest  || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${CP_VERSION} || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${VERSION} || exit 1 ; \
-		docker tag confluentinc/cp-$${component}:latest ${REPOSITORY}/cp-$${component}:${COMMIT_ID} || exit 1 ; \
+		for type in "" rpm; do \
+			DOCKER_FILE="debian/$${component}/Dockerfile"; \
+			COMPONENT_NAME=$${component}; \
+			echo $${type}; \
+			if [ "$${type}" = "rpm" ]; then \
+				echo "this an rpm"; \
```
maxzheng: The echo above should signify the type for both and so this is not needed.

debian/base/Dockerfile.rpm
```diff
@@ -0,0 +1,94 @@
...
+# ARG APT_ALLOW_UNAUTHENTICATED=false
+# #Set an env var so that it's available in derived images
+# ENV APT_ALLOW_UNAUTHENTICATED=${APT_ALLOW_UNAUTHENTICATED}
+
+MAINTAINER partner-support@confluent.io
+LABEL io.confluent.docker=true
+
+# # Python
+# # TODO: get this version
+# # ENV PYTHON_VERSION="2.7.9-1"
```
maxzheng: What do we need Python 2 for? We should remove this comment / avoid using Python 2 unless there is a hard reason that we can't use Python 3.

debian/base/Dockerfile.rpm
```diff
@@ -0,0 +1,94 @@
...
+
+RUN echo "===> clean yum caches ....." \
+    && yum clean all
+
+RUN echo "===> Installing curl wget netcat python...." \
+    && yum install -y \
+                curl \
+                wget \
+                nc \
+                python
```
maxzheng: Depending on the purpose of the echo line, if you just want to show what command is being run, it is better to use set -ex, like: 

```
RUN set -ex; \
    yum clean all; \
    yum install -y \
        curl \
        wget \
        ....
```

For the above, if you need additional comments to provide info, you can add them using # as well.

examples/cp-all-in-one/rpm-overides.yml
```diff
@@ -0,0 +1,9 @@
+# to use this: docker-compose -f docker-compose.yml -f rpm-overides.yml up
+version: '2'
+services:
```
maxzheng: How come we are only doing this for these 3 components but not others?

Makefile
norwood: i dont know? this is how the code was before i got here. i will update

Makefile
norwood: will remove

Makefile
norwood: will remove, just had this here for debugging

debian/base/Dockerfile.rpm
norwood: i agree. i was trying not to upend the entire repo in one fell swoop. if you are ok with it, then im happy to :)

debian/base/Dockerfile.rpm
norwood: just trying to remain consistent with the debs.

it also seems like changing this on 3.2.1 would be wrong. maybe that change should go on master

debian/base/Dockerfile.rpm
norwood: wilco, but again, this is just c/p from the debs. maybe a second pr where i can go through and clean all this up? i'm not sure that we should be doing cleanup in feature addition prs

examples/cp-all-in-one/rpm-overides.yml
norwood: i am

1. lazy
1. only really needed `enterprise-control-center`

debian/base/Dockerfile.rpm
norwood: actually, would you mind if we did that in another pr? its basically going to touch all the files :(

maxzheng: Of course, another pr is fine. This will help maintaining things easier in the future.

PR #258, opened 2017-05-09
we need to map moby as an extra host

norwood:
@theduderog


norwood:
i'm not 100% sure, think it has to do with `network_mode: host`, just found http://blog.yohanliyanage.com/2016/09/docker-machine-moby-name-or-service-not-known/

i usually just run `examples/cp-all-in-one` which does not set `network_mode: host`

norwood:
uggh, in which case, i should update all the other compose files as well. just verified that they too are broken :(

norwood:
k. i think this is all that need them.

PR #263, opened 2017-05-10
ST-394: we now tag our jenkins images with this label, so we get build errors  on jenkins. be more constrained

norwood:
`io.confluent.docker.build.number` may not be the greatest thing to use for this, but it is a label we only set in `cp-docker-images`

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -120,7 +120,7 @@ To get started, you can build all the CP images as follows:
 
 You can run build tests by running ``make test-build``.  Use this when you want to test the builds with a clean slate.  This deletes all images and starts from scratch.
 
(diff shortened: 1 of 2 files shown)


norwood:
ping

docs/development.rst
```diff
@@ -149,9 +149,9 @@ Make Targets
 
 Delete all images tagged with ``label=io.confluent.docker.testing=true`` :
 
-``clean-images`` 
+``clean-images``
 
-Delete all containers tagged with ``label=io.confluent.docker`` :
+Delete all containers tagged with ``label=o.confluent.docker.build.number`` :
```
maxzheng: missing i before o.confluent

PR #264, opened 2017-05-10
Make CONNECT_REST_ADVERTISED_HOST_NAME optional

ghost:
Requiring CONNECT_REST_ADVERTISED_HOST_NAME only makes sense if you are not using multiple containers not on the same Docker network.  In order to allow scaling the number of Kafka Connect instances it should be optional with the hostnames generated by Docker Compose (or similar).

Note that this doesn't currently build because of an unrelated error around the updating Debian step of the build process.  I don't know enough Debian to easily fix that.

### DIFF
--- a/debian/kafka-connect-base/include/etc/confluent/docker/configure
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/configure
@@ -28,9 +28,6 @@ dub ensure CONNECT_KEY_CONVERTER
 dub ensure CONNECT_VALUE_CONVERTER
 dub ensure CONNECT_INTERNAL_KEY_CONVERTER
 dub ensure CONNECT_INTERNAL_VALUE_CONVERTER
-# This is required to avoid config bugs. You should set this to a value that is
-# resolvable by all containers.
-dub ensure CONNECT_REST_ADVERTISED_HOST_NAME
 
 # Default to 8083, which matches the mesos-overrides. This is here in case we extend the containers to remove the mesos overrides.
 if [ -z "$CONNECT_REST_PORT" ]; then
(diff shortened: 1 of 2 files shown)


WheresAlice:
[clabot:check]

WheresAlice:
I managed to apply this to the v3.2.1 branch and it builds and works for my use-case.

rhauch:
@WheresAlice wrote:

> Requiring CONNECT_REST_ADVERTISED_HOST_NAME only makes sense if you are not using multiple containers not on the same Docker network.

Yes, you're correct that on Docker Machine and Compose it's technically not necessary, but there are lots of other environments (e.g., Kubernetes, OpenShift, DCOS, ?) where it will be required to properly notify other workers in the cluster how to contact this worker. Removing this check makes it more likely that connectivity fails in these other environments. Thoughts, @ewencp?

Isn't a simple workaround when using a single Docker network just to manually set this to the name of the container?

arrawatia:
@rhauch That's correct. There is no sensible default for `CONNECT_REST_ADVERTISED_HOST_NAME` that works well for all deployment scenarios. It also does not work if you use host networking with Docker Compose. 

The workaround you mentioned will work fine. This is how the tests for connect on host networking work. See : https://github.com/confluentinc/cp-docker-images/blob/v3.2.1/tests/fixtures/debian/kafka-connect/distributed-single-node.yml#L93

WheresAlice:
Just because many systems want you to pass in an advertised hostname doesn't mean that all do.

Removing the check allows this to run in docker-compose (and Rancher) and make use of the `scale` command to scale up the number of running containers.  With this check in place, I have to manually name and configure each instance, which means editing the docker-compose.yml file every time I want to change the scaling.

vfoucault:
Does Anyone working on this ?
maybe setting `CONNECT_REST_ADVERTISED_HOST_NAME` default value to `$HOSTNAME` worth the shot ?

I definitively need this.

PR #267, opened 2017-05-12
We are missing the new metrics config

rnpridgeon:
There does not appear to be a way to change the replication factor for the metrics topic. This is problematic if you plan to start a cluster with less than 3 brokers


theduderog:
retest this please

PR #274, opened 2017-05-26
Quick patch for minor DC/OS issue

dbtucker:
The deployment of REST Proxy in DC/OS did not set the "id" property in kafka-rest.properties.

The quick solution here is to extract a unique string from the Mesos environment to use for
that value. This will ONLY affect DC/OS deployments.

Please back-patch into 3.2.1 builds if possible.


norwood:
if you want this on 3.2.1 you should change this pr to point at v3.2.1

dbtucker:
Will re-target to specific releases.

PR #275, opened 2017-05-30
Allow accessing the Kafka broker via localhost:9092 from your Docker host machine

miguno:
This allows host-to-container communication (e.g. Mac laptop to running `kafka` container) via `localhost:9092`, in addition to inter-container communication to the Kafka broker via `kafka:29092`

**When merging into downstream branches**:  When merging into `v3.2.1` and further branches, this PR needs to add the following Docker setting to all containers:

```
     extra_hosts:
       - "moby:127.0.0.1"
```
I can provide this in a separate PR against `v3.2.1`, but I suppose we want to follow the normal merge-downwards approach...?


aayars:
@miguno I don't quite understand the special merging instructions. Can you make sure your changes are merged forward to master after committing? There are 3.3.x and 3.4.x branches now.

PR #279, opened 2017-06-21
Use default publish.ms in metrics reporter

xvrl:
This creates unncessary load in C3 and is a source of many troubles.


blueedgenick:
holy cow. can we see how that got in there in the first place ?

theduderog:
@blueedgenick  that was me :)

theduderog:
BTW, what's a reasonable value?

xvrl:
@theduderog the default (15s) is plenty enough, otherwise we're hammering the metrics aggregation topology when there are lots of metrics (e.g. when you have hundreds of topics)

xvrl:
@ijuma good point, I don't know @theduderog @arrawatia any ideas?

theduderog:
no idea

arrawatia:
I don't recall either. :)

PR #282, opened 2017-06-26
Use MESOS_TASK_ID to set "id" property in kafka-rest.properties.

dbtucker:
Minor fix for DC/OS environments to avoid multiple REST proxy instances with the same REST_ID being deployed.


dbtucker:
Thanks for the quick check, Sumit.  

Who needs to merge this in ?   I can't figure out what the issue is with the "All checks have failed" warning from Jenkins ?

PR #284, opened 2017-06-29
SNAPSHOT support for cp-docker-images 3.3.x

aayars:
Not proud of how settings.xml is handled here, open to suggestions. settings.xml is needed to point maven at nexus, since that's where the required SNAPSHOT JARs live.


Makefile
```diff
@@ -96,6 +104,17 @@ push-public: clean build-debian
 		docker push ${REPOSITORY}/cp-$${component}:${CP_VERSION} || exit 1; \
   done
 
+push-nexus:
```
maxzheng: What's the push-private used for? Feels like we should be using that for nexus by defininig DOCKER_REMOTE_REPOSITORY in env to avoid having to put nexus repo info in here as this is available for public use?

Makefile
```diff
@@ -96,6 +104,17 @@ push-public: clean build-debian
...
+	for component in ${COMPONENTS} ; do \
+		echo "\n Pushing cp-$${component}  \n==========================================\n "; \
+		docker tag ${REPOSITORY}/cp-$${component}:latest docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:latest || exit 1; \
+		docker tag ${REPOSITORY}/cp-$${component}:${VERSION} docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:${VERSION} || exit 1; \
+		docker tag ${REPOSITORY}/cp-$${component}:${CP_VERSION} docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:${CP_VERSION} || exit 1; \
+		docker push docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:latest || exit 1; \
+		docker push docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:${VERSION} || exit 1; \
+		docker push docker.confluent.io:5000/${REPOSITORY}/cp-$${component}:${CP_VERSION} || exit 1; \
```
maxzheng: Are we ok with letting people know about docker.confluent.io:5000? It is secured, but not sure about the etiquette of having internal references in public repos -- maybe @ewencp can chime in here.

Makefile
```diff
@@ -1,13 +1,21 @@
...
 MYSQL_DRIVER_VERSION := 5.1.39
 
-CONFLUENT_DEB_REPO := http://packages.confluent.io
-CONFLUENT_RPM_REPO := http://packages.confluent.io
-APT_ALLOW_UNAUTHENTICATED := false
+# CONFLUENT_DEB_REPO := http://packages.confluent.io
+CONFLUENT_DEB_REPO := https://s3-us-west-2.amazonaws.com/jenkins-confluent-packages/packaging-3.3.x/35
+
+# CONFLUENT_RPM_REPO := http://packages.confluent.io
+CONFLUENT_RPM_REPO := https://s3-us-west-2.amazonaws.com/jenkins-confluent-packages/packaging-3.3.x/35
```
maxzheng: Ok with exposing non-public urls in public repo?

debian/base/Dockerfile
```diff
@@ -37,11 +40,12 @@ ENV SCALA_VERSION="2.11"
...
 ENV CONFLUENT_MINOR_VERSION="3"
 
-ENV CONFLUENT_PATCH_VERSION="0"
+ENV CONFLUENT_PATCH_VERSION="0~SNAPSHOT"
 ENV CONFLUENT_VERSION="$CONFLUENT_MAJOR_VERSION.$CONFLUENT_MINOR_VERSION.$CONFLUENT_PATCH_VERSION"
 ENV CONFLUENT_DEB_VERSION="1"
 
-ENV KAFKA_VERSION="0.11.0.0-SNAPSHOT"
+# Add ~SNAPSHOT (with tilde) for internal releases
+ENV KAFKA_VERSION="0.11.0.0~SNAPSHOT"
```
maxzheng: After 3.3.0 is released, we will need to manually update all the files with different repo urls and also remove snapshot versions from here. If we can wrap everything in scripts to automatically change based on the git branch to produce release vs snapshots, it would be better. 

We have too many manual stuff that we have to do for each release / version, I think we need to start prioritizing on writing automated scripts as we build features rather than waiting until later. So I would recommend doing that now if possible.

debian/base/Dockerfile.rpm
```diff
@@ -36,12 +38,13 @@ ENV PYTHON_PIP_VERSION="9.0.1"
 # Confluent
 ENV SCALA_VERSION="2.11"
 ENV CONFLUENT_MAJOR_VERSION="3"
-ENV CONFLUENT_MINOR_VERSION="2"
+ENV CONFLUENT_MINOR_VERSION="3"
 
-ENV CONFLUENT_PATCH_VERSION="1"
+ENV CONFLUENT_PATCH_VERSION="0"
```
maxzheng: Wrap building of this Dockerfile in a script that automatically extracts from git branch and sets these as env vars.

debian/base/Dockerfile.rpm
```diff
@@ -36,12 +38,13 @@ ENV PYTHON_PIP_VERSION="9.0.1"
...
 ENV CONFLUENT_VERSION="$CONFLUENT_MAJOR_VERSION.$CONFLUENT_MINOR_VERSION.$CONFLUENT_PATCH_VERSION"
 
-ENV KAFKA_VERSION="0.10.2.1"
+# Add -0.1.SNAPSHOT for internal releases.
+ENV KAFKA_VERSION="0.11.0.0-0.1.SNAPSHOT"
```
maxzheng: Some things like this may be a bit hard to script. Oh well. For anything that we can't automate, let's add to a README though so someone can go thru a list next time.

debian/enterprise-kafka/Dockerfile.rpm
```diff
@@ -30,7 +30,7 @@ EXPOSE 9092
 
 RUN echo "===> installing confluent-support-metrics ..." \
     && yum -q -y update \
-    && yum install -y \
+    && yum install -y --nogpgcheck \
```
maxzheng: Stuff like this needs to be automatically toggled from a script via env otherwise it is easy to forget to change

debian/base/Dockerfile
aayars: +1. I wouldn't seriously merge this, it's too awful. We do need to come up with a wrapper to sort out the differences between build types.

Makefile
miguno: Yeah, same question.  Perhaps use an env variable instead?

debian/enterprise-kafka/Dockerfile.rpm
miguno: Yeah, this would be great.  Same for switching between public repo urls and snapshot/private ones.

Makefile
theduderog: I had the same issue so I put in a `local.make` [file that doesn't get check into this repo](https://github.com/confluentinc/cp-docker-images/blob/3.2.x/Makefile#L13-L14). 

In that file, you can override private vars:

CONFLUENT_DEB_REPO := http://s3-us-west-2.amazonaws.com/jenkins-confluent-packages/packaging-trunk/${BUILD_NUMBER}
DOCKER_REMOTE_REPOSITORY=067306319579.dkr.ecr.us-west-2.amazonaws.com

Makefile
```diff
@@ -1,13 +1,21 @@
...
+
+# Set to false for public releases
+APT_ALLOW_UNAUTHENTICATED := true
```
theduderog: this could be left `false` in the primary Makefile if we go with the local.make override approach.

Makefile
theduderog: Or they can be passed in via cmd line, `make APT_ALLOW_UNAUTHENTICATED=true`

Makefile
theduderog: Maybe that part should be a variable

debian/base/Dockerfile
```diff
@@ -20,8 +20,11 @@ LABEL io.confluent.docker.git.id=$COMMIT_ID
 ARG BUILD_NUMBER=-1
 LABEL io.confluent.docker.build.number=$BUILD_NUMBER
 
-ARG CONFLUENT_DEB_REPO="http://packages.confluent.io"
-ARG APT_ALLOW_UNAUTHENTICATED=false
+# ARG CONFLUENT_DEB_REPO="http://packages.confluent.io"
+ARG CONFLUENT_DEB_REPO="https://s3-us-west-2.amazonaws.com/jenkins-confluent-packages/packaging-3.3.x/35"
```
theduderog: I think maybe these build args [do not need default values](https://docs.docker.com/engine/reference/builder/#arg) so they could just be declared as `ARG CONFLUENT_DEB_REPO` and leave all the values in the Makefile

debian/base/Dockerfile
theduderog: I'd suggest making all the params that vary across releases build args, not setting any defaults for them in the Dockerfiles, and use the Makefile as you only place you need to change them.

debian/base/Dockerfile.rpm
```diff
@@ -72,21 +75,24 @@ RUN echo "Installing Zulu OpenJDK ${ZULU_OPENJDK_VERSION}" \
     && yum -q -y update \
     && yum -q -y install zulu-${ZULU_OPENJDK_VERSION}
 
-RUN echo "===> Adding confluent repository...${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}" \
-    && rpm --import ${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}/archive.key
+# Comment this out for unsigned SNAPSHOT releases
+# RUN echo "===> Adding confluent repository...${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}" \
+#    && rpm --import ${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}/archive.key
```
theduderog: I thought this used to be conditionalized based on the `APT_ALLOW_UNAUTHENTICATED` env var

debian/enterprise-kafka/Dockerfile.rpm
theduderog: The `APT_ALLOW_UNAUTHENTICATED` var was intended for this but perhaps is too debian specific.

debian/base/Dockerfile
theduderog: @ewencp These are build args which do not require the default values being set here so we can just remove them and then there's nothing to update for each release

Makefile
```diff
@@ -1,13 +1,21 @@
...
 
 COMPONENTS := base zookeeper kafka kafka-rest schema-registry kafka-connect-base kafka-connect enterprise-control-center kafkacat enterprise-replicator enterprise-kafka kafka-streams-examples
 COMMIT_ID := $(shell git rev-parse --short HEAD)
 MYSQL_DRIVER_VERSION := 5.1.39
 
-CONFLUENT_DEB_REPO := http://packages.confluent.io
-CONFLUENT_RPM_REPO := http://packages.confluent.io
-APT_ALLOW_UNAUTHENTICATED := false
+# CONFLUENT_DEB_REPO := http://packages.confluent.io
+CONFLUENT_DEB_REPO := https://s3-us-west-2.amazonaws.com/jenkins-confluent-packages/packaging-3.3.x/35
```
theduderog: You can currently pass them into make via command line args or a local.make file.

Makefile
ewencp: 👍  to just making it a variable. That's also nice for anyone that may be customizing the images themselves and want to publish to their own repo.

PR #285, opened 2017-06-30
Fix link in logging docs.

ewencp:


ewencp:
ping. it looks like currently we wouldn't expect the 3.3.x build to work since it seems like the Kafka version is probably out of date since 0.11.0.0 was released.

PR #292, opened 2017-07-13
SNAPSHOT and RC support via local Makefiles

aayars:
This time with feeling.

The cp-docker-images-overlay repo contains local.make files for snapshot and rc support, and a script to help drive the process.


Makefile
```diff
@@ -1,18 +1,34 @@
-BUILD_NUMBER := 1
-CP_VERSION := 3.3.0-SNAPSHOT
+# You can override vars like REPOSITORY in a local.make file
+-include local.make
```
maxzheng: It works, but will leave it to other with more experience with make to comment naming and best practice in including other Makefiles.

debian/base/Dockerfile.rpm
```diff
@@ -72,22 +78,35 @@ RUN echo "Installing Zulu OpenJDK ${ZULU_OPENJDK_VERSION}" \
     && yum -q -y update \
     && yum -q -y install zulu-${ZULU_OPENJDK_VERSION}
 
-RUN echo "===> Adding confluent repository...${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}" \
-    && rpm --import ${CONFLUENT_RPM_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}/archive.key
-
-RUN printf "[Confluent.dist] \n\
+RUN echo "===> Adding confluent repository...${CONFLUENT_PACKAGES_REPO}/rpm/${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}" && \
```
maxzheng: You might want to put all this in a bash file, and then just have RUN call the bash file to avoid having to do all the escapes and also improve readability. Also ok if you want to leave it for now.

debian/kafka-streams-examples/Dockerfile
```diff
@@ -50,6 +50,8 @@ EXPOSE $KAFKA_MUSIC_APP_REST_PORT
 # base image that supports it
 ENV LANG="C.UTF-8"
 
+COPY settings.xml /root/.m2/settings.xml
```
maxzheng: Where is settings.xml from?

debian/base/Dockerfile.rpm
aayars: +1

debian/kafka-streams-examples/Dockerfile
aayars: This your Maven settings file. For CI jobs, this is a managed file.

aayars: The CLI in overlay takes this as an arg.

aayars: Just realized it will not be there for public builds, so this needs to be made conditional. Somehow.

README.md
```diff
@@ -5,21 +5,13 @@ Docker images for deploying and running the Confluent Platform.  The images are
 
 Full documentation for using the images can be found [here](http://docs.confluent.io/current/cp-docker-images/docs/intro.html).
 
-# Important Caveat - Images Not Supported for Docker for Mac
```
ewencp: Was this removed intentionally as part of these commits? Seems unrelated to the other changes in the PR?

Makefile
ewencp: And is this something we really want or should we just have them overridden via envvars?

Makefile
```diff
@@ -1,18 +1,36 @@
...
+
+# Bump this on subsequent build, reset on new version or public release. Inherit from env for CI builds.
+BUILD_NUMBER ?= 1
+
+CONFLUENT_MAJOR_VERSION ?= 3
+CONFLUENT_MINOR_VERSION ?= 3
+CONFLUENT_PATCH_VERSION ?= 0
+
+CONFLUENT_VERSION ?= ${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}.${CONFLUENT_PATCH_VERSION}
```
ewencp: Where does `SNAPSHOT` get added? This seems relevant since below here you define `VERSION` which is going to be used as the docker image tag. Or is the expectation that you override all of `CONFLUENT_MAJOR_VERSION`, `CONFLUENT_MINOR_VERSION`, `CONFLUENT_PATCH_VERSION`, `CONFLUENT_VERSION`, and `KAFKA_VERSION` if you override anything? Would an empty-by-default but optional `SNAPSHOT` or `SUFFIX` or something like that to cover `SNAPSHOT`, `alpha`, `beta`, etc make sense here?

Makefile
```diff
@@ -32,14 +50,16 @@ debian/base/include/etc/confluent/docker/docker-utils.jar:
...
-	&& cp target/docker-utils-${CP_VERSION}-jar-with-dependencies.jar ../debian/base/include/etc/confluent/docker/docker-utils.jar \
+	&& cp target/docker-utils-${CONFLUENT_VERSION}${CONFLUENT_MVN_LABEL}-jar-with-dependencies.jar ../debian/base/include/etc/confluent/docker/docker-utils.jar \
 	&& cd -
 
 build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
 	for component in ${COMPONENTS} ; do \
 		echo "\n\nBuilding $${component} \n==========================================\n " ; \
 		if [ "$${component}" = "base" ]; then \
-			BUILD_ARGS="--build-arg APT_ALLOW_UNAUTHENTICATED=${APT_ALLOW_UNAUTHENTICATED} --build-arg CONFLUENT_DEB_REPO=${CONFLUENT_DEB_REPO}" ; \
+			BUILD_ARGS="--build-arg ALLOW_UNSIGNED=${ALLOW_UNSIGNED} --build-arg CONFLUENT_PACKAGES_REPO=${CONFLUENT_PACKAGES_REPO} --build-arg CONFLUENT_MVN_LABEL=${CONFLUENT_MVN_LABEL}"; \
```
ewencp: This customization seems a bit odd -- aren't all these things just build constants that we could pass to all the dockerfiles? Or does it complain if a build arg isn't used?

debian/base/Dockerfile
```diff
@@ -20,10 +20,11 @@ LABEL io.confluent.docker.git.id=$COMMIT_ID
...
 LABEL io.confluent.docker.build.number=$BUILD_NUMBER
 
-ARG CONFLUENT_DEB_REPO="http://packages.confluent.io"
-ARG APT_ALLOW_UNAUTHENTICATED=false
+ARG CONFLUENT_PACKAGES_REPO=$CONFLUENT_PACKAGES_REPO
+
+ARG ALLOW_UNSIGNED=false
 #Set an env var so that it's available in derived images
-ENV APT_ALLOW_UNAUTHENTICATED=${APT_ALLOW_UNAUTHENTICATED}
+ENV ALLOW_UNSIGNED=$ALLOW_UNSIGNED
```
ewencp: Not sure how we feel about compatibility of these images, but changing exported ENV vars might break downstream users of the variable.

debian/enterprise-control-center/Dockerfile
```diff
@@ -33,7 +33,7 @@ ENV CONTROL_CENTER_CONFIG_DIR=/etc/confluent-${COMPONENT}
 EXPOSE 9021
 
 RUN echo "===> Installing ${COMPONENT}..." \
-    && apt-get update && apt-get install -y confluent-${COMPONENT}=${CONFLUENT_VERSION}-${CONFLUENT_DEB_VERSION} \
+    && apt-get update && apt-get install -y confluent-${COMPONENT}=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
```
ewencp: What does `CONFLUENT_PLATFORM_LABEL` refer to exactly? What form should it take? I see that it can be set to either `CONFLUENT_RPM_LABEL` or `CONFLUENT_DEB_LABEL`, but I can't see what those would be set to.

Makefile
```diff
@@ -49,11 +69,14 @@ build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
 			if [ "$${type}" = "rpm" ]; then \
 				COMPONENT_NAME="rpm-$${component}"; \
 				DOCKER_FILE="$${DOCKER_FILE}.rpm"; \
+				CONFLUENT_PLATFORM_LABEL=${CONFLUENT_RPM_LABEL}; \
+			else \
+				CONFLUENT_PLATFORM_LABEL=${CONFLUENT_DEB_LABEL}; \
 			fi; \
 			if [ -a "$${DOCKER_FILE}" ]; then \
-				docker build --build-arg COMMIT_ID=${COMMIT_ID} --build-arg BUILD_NUMBER=${BUILD_NUMBER} $${BUILD_ARGS} -t ${REPOSITORY}/cp-$${COMPONENT_NAME}:latest -f $${DOCKER_FILE} debian/$${component} || exit 1 ; \
+				docker build --build-arg KAFKA_VERSION=${KAFKA_VERSION} --build-arg CONFLUENT_PLATFORM_LABEL=$${CONFLUENT_PLATFORM_LABEL} --build-arg CONFLUENT_MAJOR_VERSION=${CONFLUENT_MAJOR_VERSION} --build-arg CONFLUENT_MINOR_VERSION=${CONFLUENT_MINOR_VERSION} --build-arg CONFLUENT_PATCH_VERSION=${CONFLUENT_PATCH_VERSION} --build-arg COMMIT_ID=${COMMIT_ID} --build-arg BUILD_NUMBER=${BUILD_NUMBER} $${BUILD_ARGS} -t ${REPOSITORY}/cp-$${COMPONENT_NAME}:latest -f $${DOCKER_FILE} debian/$${component} || exit 1 ; \
```
ewencp: There's an inconsistent number of `$` or `$$` on this line. Some build args have a `$$` while most have `$`. Is that correct or am I missing some subtleties of the escaping that's happening here.

debian/zookeeper/Dockerfile.rpm
```diff
@@ -29,7 +29,11 @@ ENV COMPONENT=zookeeper
 
 RUN echo "===> installing ${COMPONENT}..." \
     && yum -q -y update \
-    && yum install -y confluent-kafka-${SCALA_VERSION}-${KAFKA_VERSION} \
+    && if [ "x${ALLOW_UNSIGNED}" = "xtrue" ]; then \
+        yum install -y --nogpgcheck confluent-kafka-${SCALA_VERSION}-${KAFKA_VERSION}${CONFLUENT_PLATFORM_LABEL}; \
```
ewencp: Seems like we could avoid most of these additions/variants if we just setup the yum repo config with the flag set correctly. The setup of the file is a bit messy currently, but if it's going to get refactored anyway, might be better to have that localized to one place.

Makefile
aayars: You're right that it should be in the label. I've updated this PR to do so.

For other cases, SNAPSHOT will be sent as the value of CONFLUENT_MVN_LABEL, empty for non-CI builds (but usable for special occasions if we need it)

Makefile
aayars: That's a good question. We would put it all in the wrapper script.

As it is, though, it's still easy to do builds without the wrapper script. It might be nice to keep it that way.

theduderog: One of the use cases for this is people building our images and pushing to their own repos.  Then could set their overrides in a single place and even commit in version control without having to set ENV vars all the time.  I personally hate ENV vars for lack of reproducibility.

Makefile
aayars: We could just sent it to all of them, I originally was doing so. There are already several warning for unused variables spammed to each component. I was just trying to not spam another one everywhere, since it was only needed for streams examples.

README.md
aayars: It is unrelated, I thought I would try to slip it through. Can break this out into a different PR.

Makefile
aayars: Makefile variables will have one, while shell variables have double.

debian/enterprise-control-center/Dockerfile
aayars: This is overridden by local.make in our private repo, and is combined with the revision to form the semver label for the target platform. Debian snapshots end up looking like "~SNAPSHOT-1", while RPMs look like ".0.1.SNAPSHOT". For non-snapshot cases it's just empty.

debian/zookeeper/Dockerfile.rpm
aayars: Not sure I follow -- I think we'd need to start signing snapshot packages with a bogus certificate here.

debian/base/Dockerfile
aayars: Hmm. Maybe we can provide a changelist?

README.md
ewencp: Not really a big deal, I'm just surprised because I didn't think the issues with Docker for Mac had gone away.

Makefile
ewencp: I'm fine avoiding the spam, I wasn't sure if it was going to generate any.

Makefile
ewencp: I think my confusion is from where we don't include it in the version that isn't a case where we need only a subset of the version (e.g. when we only need the `${CONFLUENT_MAJOR_VERSION}.${CONFLUENT_MINOR_VERSION}` and when those labels would differ between all those components. Is there sometime other than `-SNAPSHOT` where these labels are something different?

debian/base/Dockerfile
ewencp: I just don't know that we even have a policy wrt this repo yet, or how it's been treated for past releases. /cc @samjhecht

debian/enterprise-control-center/Dockerfile
ewencp: Ok, this was the explanation I was looking for. I guess it's unlikely anybody else does snapshot builds, but just sticking that as a comment above the variables in the `Makefile` would really clarify what they are for.

debian/zookeeper/Dockerfile.rpm
ewencp: You can set `gpgcheck=1` in the `confluent.repo` file and it won't bother with anything gpg-related.

Makefile
aayars: Including -SNAPSHOT doesn't work for deb or rpm commands. It ended up working out best for all cases to provide a base and pin the right suffix on as needed.

debian/zookeeper/Dockerfile.rpm
aayars: I thought I was doing this (1 for enabled) -- Apparently the unsigned repo config needs gpgcheck=0 explicitly, or it will inherit a default of 1. Confirmed that this works when setting it to 0 in the repo config.

debian/base/Dockerfile
aayars: I think we need the flexibility to change anything which isn't part of a documented public API. Not being able to change the stuff in these Makefiles and Dockerfiles will really bind us.

Documented env variables are at http://docs.confluent.io/current/cp-docker-images/docs/development.html#extending-the-docker-images

ewencp: Sure, and we might say this case is fine since these should be built off of our signed repos for release.

That said, I'm not sure what docs you're referring to. We have some examples in that code, but there are plenty more env variables in each of the dockerfiles derived from this one that people rely on and are not explicitly documented afaict.

aayars:
Retest this please

aayars:
Retest this please

aayars:
Retest this please

aayars:
Retest this please

aayars:
The SNAPSHOT PR builder isn't going to work, due to nexus settings.

PR #299, opened 2017-07-21
move build-debian step into a script

aayars:


bin/build-debian
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
maxzheng: Can u rename to build-debian.sh? That's the convention everywhere. 

Also the sheshang should be: #!/usr/bin/env bash
It's more portable.

maxzheng (approved):

PR #300, opened 2017-07-27
added support for CONTROL_CENTER_OPTS env

coughman:
this allows specifying JAAS and other things to start control center


coughman:
turns out this is not needed

PR #301, opened 2017-07-28
setting offsets topic replication factor to 1

bbejeck:


bbejeck:
ping @ybyzek @miguno

Tested locally agains internal nexus images.

bbejeck:
merged to 3.4.x and master with pint.

PR #309, opened 2017-08-14
CC-799: Update Connect docker quickstart.

kkonstantine:


kkonstantine:
This PR updates the Connect quickstart with the necessary arguments to the connect worker, and removes the need to login to docker-machine, which was broken, since the directories and files were not created at the correct host, location and with the right permissions.

docs/quickstart.rst
```diff
@@ -207,19 +211,19 @@ Now that we have Kafka and Zookeeper up and running, we can deploy some of the o
...
 
-    docker run -it --net=host --rm confluentinc/cp-schema-registry:3.3.0-SNAPSHOT bash
+    $ docker run -it --net=host --rm confluentinc/cp-schema-registry:3.3.0-SNAPSHOT bash
 
   Direct the utility at the local Kafka cluster, tell it to write to the topic ``bar``, read each line of input as an Avro message, validate the schema against the Schema Registry at the specified URL, and finally indicate the format of the data.
 
   .. sourcecode:: bash
 
-    /usr/bin/kafka-avro-console-producer \
+    # /usr/bin/kafka-avro-console-producer \
```
rhauch: Should this be `$` instead of `#`?

docs/quickstart.rst
```diff
@@ -534,28 +538,25 @@ Now you should verify that the topics are created before moving on:
...
        --net=host \
        --rm \
        confluentinc/cp-kafka:3.3.0-SNAPSHOT \
        kafka-topics --describe --zookeeper localhost:32181
 
-For this example, we'll create a FileSourceConnector, a FileSinkConnector and directories for storing the input and output files. If you are running Docker Machine then you will need to SSH into the VM to run these commands by running ``docker-machine ssh <your machine name>``.
+For this example, we'll create a FileSourceConnector, a FileSinkConnector and directories for storing the input and output files.
 
-  First, let's create the directory where we'll store the input and output data files.  Remember, we must do this within the Docker Host.
+  First, let's create the directory where we'll store the input and output data files.
```
rhauch: Since this is the first time we're running `docker exec` (?), perhaps it's worth adding something like "in the running container."

docs/quickstart.rst
```diff
@@ -493,23 +497,23 @@ First, let's start up a container with Kafka Connect.  Connect stores all its st
```
rhauch: We no longer have to create these topics by hand, so should we remove these steps to keep things as simple as possible?

docs/quickstart.rst
```diff
@@ -566,20 +567,24 @@ For this example, we'll create a FileSourceConnector, a FileSinkConnector and di
...
         -e CONNECT_VALUE_CONVERTER="org.apache.kafka.connect.json.JsonConverter" \
         -e CONNECT_INTERNAL_KEY_CONVERTER="org.apache.kafka.connect.json.JsonConverter" \
         -e CONNECT_INTERNAL_VALUE_CONVERTER="org.apache.kafka.connect.json.JsonConverter" \
         -e CONNECT_REST_ADVERTISED_HOST_NAME="localhost" \
         -e CONNECT_LOG4J_ROOT_LOGLEVEL=DEBUG \
+        -e CONNECT_LOG4J_LOGGERS=org.reflections=ERROR \
         -v /tmp/quickstart/file:/tmp/quickstart \
         confluentinc/cp-kafka-connect:3.3.0-SNAPSHOT
 
   As you can see in the above command, we tell Connect to refer to the three topics we create in the first step of this Connect tutorial. Let's check to make sure that the Connect worker is up by running the following command to search the logs:
```
rhauch: If we're going to have Connect create the internal topics, we should change the first sentence to:

> As you can see in the above command, we tell Connect to use three topics to store configuration, offsets, and status, and it will create these correctly if they don't yet exist. We're only running one broker, so we do need to specify a replication factor of `1` for these topics.

docs/quickstart.rst
```diff
@@ -588,43 +593,41 @@ For this example, we'll create a FileSourceConnector, a FileSinkConnector and di
     [2016-08-25 18:25:19,665] INFO Herder started (org.apache.kafka.connect.runtime.distributed.DistributedHerder)
     [2016-08-25 18:25:19,676] INFO Kafka Connect started (org.apache.kafka.connect.runtime.Connect)
 
-  We will now create our first connector for reading a file from disk.  To do this, let's start by creating a file with some data. Again, if you are running Docker Machine then you will need to SSH into the VM to run these commands by running ``docker-machine ssh <your machine name>``. (You may also need to run the command as root).
+  We will now create our first connector for reading a file from disk. To do this, let's start by creating a file with some data:
```
rhauch: change to:

> We will now create our first connector for reading a file in the container. To do this, let's start by having the container create a file with some data:

docs/quickstart.rst
```diff
@@ -588,43 +593,41 @@ For this example, we'll create a FileSourceConnector, a FileSinkConnector and di
...
   The next step is to create the File Source connector.
 
   .. sourcecode:: bash
 
-    docker@confluent:~$: curl -X POST \
+    $ docker exec kafka-connect curl -s -X POST \
       -H "Content-Type: application/json" \
-      --data '{"name": "quickstart-file-source", "config": {"connector.class":"org.apache.kafka.connect.file.FileStreamSourceConnector", "tasks.max":"1", "topic":"quickstart-data", "file": "/tmp/quickstart/input.txt"}}' \
+      --data '{"name": "quickstart-file-source", "config": {"connector.class":"org.apache.kafka.connect.file.FileStreamSourceConnector", "tasks.max":"1", "topic":"quickstart-data", "file": "/tmp/quickstart/file/input.txt"}}' \
       http://$CONNECT_HOST:28082/connectors
```
rhauch: Okay, I understand that we probably have to run `curl` from within the container so that everybody's environment is the same, but it sure seems like the user might want to run `curl` on their local machine. Otherwise, it's like you have to first SSH into the machine in order to use Connect's REST API. Perhaps that's for a different issue.

rhauch: Or perhaps in this PR we could at add a note about why we're running `curl` from within the container.

docs/quickstart.rst
kkonstantine: Unfortunately, it's not, at least not with our current images. When we run a shell (`bash` here) it will run it as root. I skipped the most "frightening" part of the prompt that will actually be `root@confluent:/#` so that we can be both accurate about what will happen and not too distracting at this quickstart.

docs/quickstart.rst
```diff
@@ -259,13 +259,13 @@ This section describes how to deploy the REST Proxy container and then consume d
...
 
-    docker run -it --net=host --rm confluentinc/cp-schema-registry:3.3.0-SNAPSHOT bash
+    $ docker run -it --net=host --rm confluentinc/cp-schema-registry:3.3.0-SNAPSHOT bash
 
   The first step in consuming data via the REST Proxy is to create a consumer instance.
 
   .. sourcecode:: bash
 
-    curl -X POST -H "Content-Type: application/vnd.kafka.v1+json" \
+    # curl -X POST -H "Content-Type: application/vnd.kafka.v1+json" \
```
kkonstantine: Same explanation. Anything that runs after we've run `docker run ... bash` will give the user root shell.

docs/quickstart.rst
```diff
@@ -279,7 +279,7 @@ This section describes how to deploy the REST Proxy container and then consume d
 
   .. sourcecode:: bash
 
-    curl -X GET -H "Accept: application/vnd.kafka.avro.v1+json" \
+    # curl -X GET -H "Accept: application/vnd.kafka.avro.v1+json" \
```
kkonstantine: Same explanation. Anything that runs after we've run `docker run ... bash` will give the user root shell.

docs/quickstart.rst
kkonstantine: I thought it didn't hurt to leave those (after I forgot to remove them in the first place). But I agree, better to start using this ability.

docs/quickstart.rst
kkonstantine: Done

docs/quickstart.rst
kkonstantine: Ok

docs/quickstart.rst
kkonstantine: It's not great to have to grab the IP of the container node to run `curl` here. I think it's not to bad to use `docker exec` to avoid confusion.

kkonstantine:
Thanks!

PR #312, opened 2017-08-15
added client auth check 

mhowlett:
The kafka configure script enforces that a truststore has been configured if it detects an SSL listener has been configured. However whether or not this is required depends on the setting for ssl.client.auth. If it is set to none (which is the default!) - truststore is not used (and doesn't need to be configured). So this is both potentially misleading and annoying. truststore should be configured if ssl.client.auth is set to required or requested.


mhowlett:
[clabot:check]

PR #314, opened 2017-08-16
CLIENTS-513 make zk url optional in docker images for SR and Rest Proxy

mageshn:


debian/kafka-rest/include/etc/confluent/docker/ensure
```diff
@@ -32,8 +35,18 @@ then
...
-        "${KAFKA_REST_CUB_KAFKA_TIMEOUT:-40}" \
-        -z "$KAFKA_REST_ZOOKEEPER_CONNECT"
+    if [[ -n "${KAFKA_REST_ZOOKEEPER_CONNECT-}" ]]
+    then
+        cub kafka-ready \
+            "${KAFKA_REST_CUB_KAFKA_MIN_BROKERS:-1}" \
+            "${KAFKA_REST_CUB_KAFKA_TIMEOUT:-40}" \
+            -z "$KAFKA_REST_ZOOKEEPER_CONNECT"
+    fi
+    if [[ -n "${KAFKA_REST_BOOTSTRAP_SERVERS-}" ]]
```
theduderog: Seems like we should run one or the other here, not both

debian/schema-registry/include/etc/confluent/docker/ensure
```diff
@@ -32,8 +34,18 @@ then
         -b "${SCHEMA_REGISTRY_KAFKASTORE_BOOTSTRAP_SERVERS}" \
         --config /etc/"${COMPONENT}"/admin.properties
 else
-    cub kafka-ready \
-        "${SCHEMA_REGISTRY_CUB_KAFKA_MIN_BROKERS:-1}" \
-        "${SCHEMA_REGISTRY_CUB_KAFKA_TIMEOUT:-40}" \
-        -z "$SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL"
+    if [[ -n "${SCHEMA_REGISTRY_KAFKASTORE_CONNECTION_URL-}" ]]
```
theduderog: I think we should only do this on if the other one didn't happen

mageshn:
I have incorporated the comments and added tests but I haven't found a reliable way to run the build and test on my laptop. Thats the reason I didn't touch the test in the first place :). Hopefully this should work.

PR #342, opened 2017-09-26
added kafka tools to the confluent control centre image

simplesteph:
Without it, we can't run basic commands such as a kafka console consumer.

### DIFF
--- a/debian/enterprise-control-center/Dockerfile
+++ b/debian/enterprise-control-center/Dockerfile
@@ -34,7 +34,7 @@ EXPOSE 9021
 
 RUN echo "===> Installing ${COMPONENT}..." \
     && apt-get update && apt-get install -y confluent-${COMPONENT}=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
-    \
+    && apt-get install confluent-kafka-${SCALA_VERSION}=${KAFKA_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Cleaning up ..."  \
     && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/* \
     \


ewencp:
@simplesteph Not sure I understand the motivation/use case here. Why wouldn't you just run console consumer in its own container? Currently the tools aren't separated out, so installing all of Kafka is kind of heavyweight. And the `control-center-console-consumer` that is included with Control Center works fine since the Kafka jars are shipped in the Control Center directory.

simplesteph:
@ewencp this has to do with running debug within the docker image. I encountered the case where the support team asked me to do that, and I had to install the kafka tools from within. Took a bit of time to find the command, etc.

The idea is that the space tradeoff here is minimal, compared to how much one gains time by having the kafka commands from within the docker image. 

If there's a simpler ways to achieve this while having minimum space, let me know. Maybe it's a just confluent side of thing to include the scripts to launch `kafka-console-consumer` for example as the kafka binaries are already there according to you.

ewencp:
@simplesteph Not sure there's a faster way if you want the shell script wrapper for the command (you could do it via control center's run-class script, but it'd be a bit more manual). I ask about space because I've found docker users to be extremely sensitive to this tradeoff, to the point that they might even rather have separate images / derive their own from this base image to give them the tradeoff of space vs convenience.

simplesteph:
@ewencp I understand. I'll close this. 

My take on space is that Docker users get very minimal (alpine & co) when you need rapid scaling of your application. Pulling 10MB is quick, and needed if your architecture is 100 containers behind an ELB scaling up and down every hour. For longer running more stable images (such as Kafka, Control Centre, REST Proxy, etc...), space is less of an issue. The only cost is the first pull and caching. Machine to Machine networks are relatively quick too and once it's there, it's there. 

Anyway, as you said a better way to solve this problem would be to just create the `.sh` files, as the jars are there, which is a more tedious task to invest it. Maybe something the c3 product team wants to look into such as next time it's installed the `.sh` are there as well

PR #343, opened 2017-09-26
Fix: Wrong templating property and env variables for C3 License

simplesteph:


simplesteph:
@ewencp I see your point. The change should now be backwards compatible

ewencp:
LGTM, thanks for the fix!

PR #361, opened 2017-10-12
Fix title and typos

joel-hamill:


sgerrand:
👏

docs/security/security.rst
```diff
@@ -1,32 +1,36 @@
 .. _security_with_docker :
 
-Security Overview
-=================
+Docker Security
+===============
 
-Using security is optional - non-secured clusters are supported, as well as a mix of authenticated, unauthenticated, encrypted and non-encrypted clients.  The following security features are currently supported on the Confluent Platform Docker images:
+Using security is optional. Confluent Platform supports non-secured clusters and a mix of authenticated, unauthenticated,
```
edenhill: It might read better to start with the secure options and then listing the less secure in the order of security.

docs/security/security.rst
```diff
@@ -1,32 +1,36 @@
...
-~~~~~~~~~~~~~~~
+Managing secrets
+  When you enable security for the Confluent Platform, you must pass secrets (e.g., credentials, certificates, keytabs,
+  Kerberos config etc.) to the container. The images handle this by expecting the credentials to be available in the
+  secrets directory. The containers specify a Docker volume for secrets and expect the admin to map it to a directory on the host
+  which contains the required secrets.
 
-1. Managing secrets
+Running containers with arbitrary user IDs
+  The images can be run with arbitrary user IDs. This provides an additional security layer against processes achieving
```
edenhill: achieving -> gaining?

docs/security/security.rst
```diff
@@ -1,32 +1,36 @@
...
+  escalated permissions on the host node by escaping the container if there is a container engine vulnerability.
 
-  When you enable security for the Confluent Platform, you need to pass secrets (credentials, certificates, keytabs, Kerberos config etc.) to the container. The images handle this by expecting the credentials to be available in the secrets directory. We specify a docker volume for secrets and expect the admin to map it to a directory on the host which contain the required secrets.
 
-2. Running containers with arbitrary User IDs
+For details on the available security features in Confluent platform, see the `Confluent Platform Security
+Overview Documentation <http://docs.confluent.io/current/kafka/security.html>`_.
 
-  The images can be run with arbitrary User IDs. This provides an additional security layer against processes achieving escalated permissions on the host node by escaping the container if there is a container engine vulnerability.
+For a tutorials on using SSL in the Confluent Platform, see the documented tutorials on `SSL
```
edenhill: "tutorial", or no "a"

docs/security/security.rst
joel-hamill: @edenhill do you mean something like this? `Confluent Platform supports cluster encryption and authentication, including a mix of authenticated and unauthenticated, and 
encrypted and non-encrypted clients. Using security is optional.`?

edenhill: Yep, that's more positive :)

PR #365, opened 2017-10-26
use java deps from common-docker/utility-belt

aayars:
Verified cub in cp-base works with new java deps.


java/pom.xml
```diff
@@ -26,162 +31,18 @@
...
-            <version>${argparse4j.version}</version>
+            <groupId>io.confluent</groupId>
+            <artifactId>utility-belt</artifactId>
+            <version>${project.version}</version>
         </dependency>
+
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
+            <artifactId>slf4j-simple</artifactId>
```
ewencp: Is this right? Seems like it would change the logging behavior. are we using this instead of the slf4j-log4j we use everywhere else for some reason? Also, is it going to affect the logging output format, which could be a compatibility issue?

ewencp (commented):
just one question re: dependencies, but otherwise is looking good. like the other PR, 🎉 removing code 🎉  and reducing duplication is excellent.

java/pom.xml
aayars: It's weird, but this is what I arrived at after testing the images. The logger doesn't seem to work with -api alone.

I added the -api dependency to common-docker and removed it here, but that cause the image to default to a noop logger, as opposed to the official image's behavior of logging verbosely. I added the -api dependency back in directly, and logging was still busted with a clean build. I stackoverflowed as hard as I could, and saw that others solved this problem by depending on the -simple artifact instead, which seemed to work. I am guessing something was getting pulled in transitively before, but stopped after the move to common-docker.

theduderog (commented):
Awesome. Thanks.

PR #366, opened 2017-10-30
MMA-1686 setting CONTROL_CENTER_DATA_DIR is no longer required

xvrl:
* 4.0.x now defaults to /var/lib/confluent-control-center
* also remove CONTROL_CENTER_CONFIG_DIR to be more consistent with our other images

### DIFF
--- a/debian/enterprise-control-center/Dockerfile
+++ b/debian/enterprise-control-center/Dockerfile
@@ -26,8 +26,6 @@ MAINTAINER partner-support@confluent.io
 LABEL io.confluent.docker=true
 
 ENV COMPONENT=control-center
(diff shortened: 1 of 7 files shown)


rancp (approved):
🎃  🐳  LGTM

arrawatia (approved):
LGTM but @theduderog might have more to say, he created the C3 docker image.

theduderog:
You don't want this to be overridable anymore?  It's possible someone uses this to get their mount dir.

xvrl:
@theduderog it's still overridable, just not a required setting anymore. The default now matches what the docker image set it to in the first place.

theduderog (approved):
Got it.  Thanks, @xvrl  🥇

PR #372, opened 2017-11-08
MMA-1538 sane way of configuring security in c3 docker image

xvrl:


debian/enterprise-control-center/include/etc/confluent/docker/admin.properties.template
```diff
@@ -1,4 +1,14 @@
-{% set security_props = env_to_props('CONTROL_CENTER_STREAMS_PRODUCER_', '') -%}
+{% set producer_security_props = env_to_props('CONTROL_CENTER_STREAMS_PRODUCER_', '') -%}
+{% set security_props = env_to_props('CONTROL_CENTER_STREAMS_', '') -%}
 {% for name, value in security_props.iteritems() -%}
+{% if not name in producer_security_props -%}
 {{name}}={{value}}
+{% endif %}
+
+{# producer specific overrides -#}
+{% if len(producer_security_props) > 0 -%}
```
ewencp: why do we need this check?

xvrl: forgot to remove it, I had a comment in there that I didn't want to write out unless we had any overrides

debian/enterprise-control-center/include/etc/confluent/docker/admin.properties.template
```diff
@@ -1,4 +1,14 @@
-{% set security_props = env_to_props('CONTROL_CENTER_STREAMS_PRODUCER_', '') -%}
+{% set producer_security_props = env_to_props('CONTROL_CENTER_STREAMS_PRODUCER_', '') -%}
+{% set security_props = env_to_props('CONTROL_CENTER_STREAMS_', '') -%}
```
theduderog: I think you can do the same with this  `{% set kafka_props = env_to_props('CONTROL_CENTER_STREAMS_', '', exclude= producer_security_props) -%}`

xvrl:
@theduderog @ewencp mind having a look again?

tests/fixtures/debian/control-center/standalone-config.yml
```diff
@@ -79,3 +79,29 @@ services:
...
+      CONTROL_CENTER_MAIL_ENABLED: "true"
+      CONTROL_CENTER_MAIL_HOST_NAME: "foo.com"
+      CONTROL_CENTER_STREAMS_SECURITY_PROTOCOL: SASL_SS
+      CONTROL_CENTER_STREAMS_SASL_KERBEROS_SERVICE_NAME: kafka
+      CONTROL_CENTER_STREAMS_SSL_KEYSTORE_LOCATION: /path/to/keystore
+      CONTROL_CENTER_STREAMS_SSL_KEYSTORE_PASSWORD: password
+      CONTROL_CENTER_STREAMS_SSL_KEY_PASSWORD: password
+      CONTROL_CENTER_STREAMS_SSL_TRUSTSTORE_LOCATION: /path/to/truststore
+      CONTROL_CENTER_STREAMS_SSL_TRUSTSTORE_PASSWORD: password
+      CONTROL_CENTER_STREAMS_PRODUCER_SECURITY_PROTOCOL: SS
```
ewencp: presumably this is supposed to be `SSL`? and in the test as well? doesn't matter much since it's just an example, but i could imagine someone getting very frustrated while overlooking that one missing letter

xvrl: @ewencp I used the same values that are used above in the `wildcards-config` section. Maybe this is a case of `sed -i` gone wild, but the way I understood it this was probably to ensure we didn't accidentally end up with SSL for some other reason and that those were the actual env values.

ewencp: OK, I don't really understand the reasoning, but given this is all test data I guess it's fine. If we really have a reason for not using SSL, I'd doc it somewhere since I'd probably see this as a typo and just fill in something valid instead if I were updating this.

theduderog: it this is testing an override, can we make it something less confusing like `FAKE_PROTOCOL`?

tests/test_control_center.py
```diff
@@ -113,6 +113,24 @@ def test_wildcards_config(self):
...
 
+    def test_admin_props_with_producer_overrides(self):
+        output = self.cluster.run_command_on_service("security-config-with-producer-override",
+                "bash -c 'while [ ! -f /tmp/config-is-done ]; do echo waiting && sleep 1; done; echo PASS'")
+        assert "PASS" in output
+
+        admin_props = props_to_list(self.cluster.run_command_on_service("security-config-with-producer-override",
+            "cat /etc/confluent-control-center/admin.properties"))
+        admin_expected = props_to_list("""
+        security.protocol=SS
```
theduderog: use something other than `SS`

tests/fixtures/debian/control-center/standalone-config.yml
xvrl: maybe @arrawatia knows why?

theduderog: you mean why `SS` was originally used?

xvrl: @theduderog yes

theduderog: he's out for 2 more weeks.  i'm in favor of changing it for clarity.

xvrl: @theduderog looks like you added some of those in the docs here https://github.com/confluentinc/cp-docker-images/commit/7c93d2536d4d655dc99d44, any reason you used SASL_SS instead of SASL_SSL?

theduderog: lol.  doh!  probably no good reason except by mistake.

xvrl: k, fixed them all and renamed things

PR #384, opened 2017-11-30
MINOR: Fix cp version and connect plugin path

kkonstantine:


kkonstantine:
Confirmed the quickstart now works for everything, including Connect.

docs/conf.py
```diff
@@ -59,7 +59,7 @@ def setup(app):
 # The short X.Y version.
 version = '3.4'
 # The full version, including alpha/beta/rc tags.
-release = '3.4.0-SNAPSHOT'
+release = '4.0.0'
```
ewencp: if we're updating the `release`, the `version` right above should get the same treatment

ewencp: oh, i guess we probably commonly miss the ones in the local `docs/conf.py` since these configs aren't used in the aggregate docs build...

kkonstantine:
Thanks everyone!

PR #397, opened 2018-01-04
Add optional ARG overrides of required ENV params during build time

mikesparr:
Adding flexibility to the Kafka Docker image to allow `ARG` overrides of the two required environment variables.

## Example usage:
```bash
docker build -t kafka \
  --build-arg KAFKA_ZOOKEEPER_CONNECT=localhost:2181 \
  --build-arg KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://test:9092  .
```

## Test:
Add `RUN echo $KAFKA_ADVERTISED_LISTENERS` in Dockerfile after proposed edits and build using example above. The console will display the passed-in value for that env param and you can omit it in the `-e ` args during runtime as it is already set.

## Helps with orchestration (i.e. Kubernetes)
This also solves a major issue with attempted Kubernetes deployments using this image, given the unique POD IP cannot be passed dynamically in the `ENV` parameters. By adding the ability to also declare these two required values (checked at startup by configure script), this image can be used with minor customization in Kubernetes.

Example (relevant snippet) Kubernetes config:
```yaml
      containers:
      - name: cp-kafka
        imagePullPolicy: Always
        image: confluentinc/cp-kafka:3.3.1
        resources:
          requests:
            memory: "4Gi"
            cpu: 0.5
        ports:
        - containerPort: 9092
          name: server
          protocol: TCP
        - containerPort: 49999
          name: jmx
          protocol: TCP
        lifecycle:
          postStart:
            exec:
              command:
                - "/bin/sh"
                - "-c"
                - "export KAFKA_BROKER_ID=${HOSTNAME##*-}"
        env:
          - name: KAFKA_ZOOKEEPER_CONNECT
            value: zk-cs.default.svc.cluster.local:2181
          - name: KAFKA_JMX_PORT
            value: "49999"
          - name: MY_POD_IP
            valueFrom:
              fieldRef:
                fieldPath: status.podIP
        args:
          - KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://$(MY_POD_IP):9092
```

### DIFF
--- a/debian/kafka/Dockerfile
+++ b/debian/kafka/Dockerfile
@@ -23,6 +23,12 @@ LABEL io.confluent.docker.build.number=$BUILD_NUMBER
 MAINTAINER partner-support@confluent.io
 LABEL io.confluent.docker=true
 
+# allow arg override of required env params
+ARG KAFKA_ZOOKEEPER_CONNECT
+ENV KAFKA_ZOOKEEPER_CONNECT=${KAFKA_ZOOKEEPER_CONNECT}
+ARG KAFKA_ADVERTISED_LISTENERS
+ENV KAFKA_ADVERTISED_LISTENERS=${KAFKA_ADVERTISED_LISTENERS}
+
 ENV COMPONENT=kafka
 
 # primary
--- a/debian/kafka/include/etc/confluent/docker/run
+++ b/debian/kafka/include/etc/confluent/docker/run
@@ -19,6 +19,16 @@ set -o nounset \
     -o verbose \
     -o xtrace
 
+
+# Set environment values if they exist as arguments
+if [ $# -ne 0 ]; then
+  echo "===> Overriding env params with args ..."
+  for var in "$@"
+  do
+    export "$var"
+  done
+fi
+
 echo "===> ENV Variables ..."
 env | sort
 


mikesparr:
@ewencp When I squash rebased it pulled in your commit as well but hopefully since that is in docs, trivial to consider and merge. Thanks. There are dozens of stack overflow and github issues trying to find workarounds to get Kafka / ZK to work on Kubernetes using StatefulSet and without major edits to the image, this seems like the cleanest way to enable those. Others have had to build their own images, then add `initContainer` step to dynamically write a kafka properties file to get the listeners set, and then override all configs and command args in image. 

These 4 lines added to Dockerfile do not impact existing but extend possibilities to solve all the Kubernetes deployment issues.

mikesparr:
@ewencp if this passes review could we expect it to be updated in current images or would it have to wait for 4.0?

ewencp:
@mikesparr 4.0.0 is already out and tagged. if it was to get into a 4.0.1 it would have to be targeted against the 4.0.x branch. otherwise it won't show up until the 4.1.0 release (sometime late feb). someone more familiar can indicate whether it makes sense for a 4.0.1 release.

mikesparr:
@ewencp sooner the better so please let me know what I need to do to speed up acceptance/merge (assuming review approved of course).

mikesparr:
@ewencp when I attempt to change to `4.0.x` branch it tries to merge 65 commits. Obviously I only want to propose the two commits (edit to `Dockerfile` and edit to `run` script). Let me know best way forward and happy to oblige or appreciate your targeting it to soonest possible release.

ewencp:
@mikesparr usually you adjust the branch to be based off of `origin/4.0.x`. the extra commits you are seeing is because the history of your commit includes changes that exist on master but not on 4.0.x. to adjust you would start a branch against `origin/4.0.x`, cherry-pick your commits, and then force push to your `arg-env-override` branch and adjust the target on this pr. if you get stuck i can help you out with this too -- normally for forked repos like this the owners of the original repo have write access to the forked copy.

mikesparr:
@ewencp if you are willing to help move those to `4.0.x` I appreciate. Thanks for insights.

ewencp:
@mikesparr updated, double check the diff to make sure it looks right

mikesparr:
@ewencp thanks. the edit to `run` script looks good but missing the edit to `Dockerfile`. See below what it should be (adding the ARG and ENV lines: 4 total)

If I had to choose, the `run` script edits are most important to solve Kubernetes issue but the flexibility adding the ARG to Dockerfile don't impact existing deployments and add flexibility on how users declare those two required ENV params.

```

FROM confluentinc/cp-base

ARG COMMIT_ID=unknown
LABEL io.confluent.docker.git.id=$COMMIT_ID
ARG BUILD_NUMBER=-1

MAINTAINER partner-support@confluent.io
LABEL io.confluent.docker=true

# allow arg override of required env params
ARG KAFKA_ZOOKEEPER_CONNECT
ENV KAFKA_ZOOKEEPER_CONNECT=${KAFKA_ZOOKEEPER_CONNECT}
ARG KAFKA_ADVERTISED_LISTENERS
ENV KAFKA_ADVERTISED_LISTENERS=${KAFKA_ADVERTISED_LISTENERS}

ENV COMPONENT=kafka

# primary
EXPOSE 9092

RUN echo "===> installing ${COMPONENT}..." \
    && apt-get update && apt-get install -y confluent-kafka-${SCALA_VERSION}=${KAFKA_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
    \
    && echo "===> clean up ..."  \
    && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/* \
    \
		&& echo "===> Setting up ${COMPONENT} dirs..." \
   	&& mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
   	&& chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets
```

ewencp:
Sorry, missed that because the commit had been squashed into a separate merge commit. See if that looks right now.

mikesparr:
Perfect. If we can get this in a recent release it should solve Kubernetes' deployment workarounds and I'll pitch this image to k8s users as the default Kafka image. Thanks for helping and considering!

cmccabe:
I don't understand the context here, but the change itself seems straightforward.  Passing through those variables as docker args, then environment variables

mikesparr:
@cmccabe correct. This PR adds flexibility to how a user can declare ENV vars with this image, instead of just the `-e flag`. The issue lies in the fact that some container orchestration systems (i.e. Kubernetes) do not provide a way to define some of the required vars dynamically. 

## spin up a 3-pod cluster using StatefulSet
 * names them kafka-1, kafka-2, kafka-3 automatically
 * generates IP automatically

```yaml
   env:
   - name: KAFKA_BROKER_ID
     value: cannot get ID dynamically
   - name: KAFKA_ADVERTISED_LISTENERS
     value: cannot concat pod IP with PLAINTEXT://(pod ip):9092
```

There is a manifest type called `StatefulSet` (replaced PetSet) that guarantees an ordinally-index cluster that are created and destroyed and named consistently. This caters to a cluster for databases, queues, etc. but since they (and Kafka) require the IDs and discovery URLs dynamically, people have resorted to "hacks" to get Kafka, ZK, etc. to run. Currently they have to add a second `initContainer` that will mount to the same volume and write a configuration file in the directory the db or queue image will look at to get the Pod's unique ID and hostname.

## With this PR adding ability to declare args in Kubernetes manifest
```yaml
  env:
    - name: KAFKA_ZOOKEEPER_CONNECT
      value: zk-cs.default.svc.cluster.local:2181
    - name: KAFKA_JMX_PORT
      value: "49999"
    - name: POD_IP
      valueFrom:
        fieldRef:
          fieldPath: status.podIP
   args:
   - "KAFKA_BROKER_ID=${HOSTNAME##*-}"
   - "KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://$(POD_IP):9092"
```

This PR solves these workarounds by leveraging the `args` object which will append args to the `CMD` of any image. If we can pass in the required ENV vars as `args` then Kubernetes can use Confluent's image for Kafka without needing the hacks. This is a more elegant solution and does not interfere with existing users, and makes the image more flexible.

cmccabe:
LGTM

gAmUssA:
@mikesparr are you running Kafka in k8s with `cp-docker-images`?

mikesparr:
@gAmUssA yes. For ZK I used a different image that had those "hacks" I mentioned but given I wanted to stay close to CP for our stream pipeline stack, I stubbornly tried to get CP Kafka to work. It runs fine but I learned after testing some client producer/consumers that the advertised URL from ENV wouldn't work, thus request for this PR. I did build an edited image and test using this PR and does solve issues but awaiting official image as we prepare a new microservice stack for stream.

If we applied this same update to the `auth` script in the ZK image, then I'm confident that too could run in k8s without the hacks. I saw another PR #403 and suggested instead they follow this pattern and use the `args` parameter in k8s manifest to dynamically pass in required IDs. 

I haven't yet looked into the rest api, schema registry images so unsure if we'll have issues there too. Goal was to use AVRO and schema registry, thus determined to get CP kafka image working first. ;-) Thanks for considering my PR.

gAmUssA:
@mikesparr it was my PR :) Let me know how's your k8s progress going?

mikesparr:
@ewencp @cmccabe how often do you push images to Docker hub? I'm wondering when I can test the public image. Thanks!

 * reference: https://hub.docker.com/r/confluentinc/cp-kafka/tags/

ewencp:
@mikesparr They're pushed to docker hub on releases. This one went to 4.0.x, so 4.0.1 will probably be the next release with this. That's planned for sometime around end of Jan. After that a 4.1 should be out sometime late Feb, which would also include this change.

gAmUssA:
@mikesparr you can use the following gist for local snapshot build https://gist.github.com/gAmUssA/8416b7db600874fecb91a1557b738bbf

- create `local.make` in the root of `cp-docker-mages`
- start the build with `myLocalBuild.sh` 
- to speed up the build you can disable somу components in Makefile (`COMPONENTS` var) 

1 projectsconfluentcp-docker-images git-remote-https 2018-01-17 19-50-05

jmcristobal:
Hi
We are currently setting CP in  k8s and this PR will be really helpful to avoid all mentioned hacks. 
Do you have any update on the release plan for 4.0.1 version?

ron1:
FYI, it appears the Kubernetes Kafka Helm Chart statefulset template (https://github.com/kubernetes/charts/blob/master/incubator/kafka/templates/statefulset.yaml) implements its own workaround that does not use args and is therefore not dependent upon this patch. The relevant snippet is included below:

command:
- sh
- -exc
  - |
    export KAFKA_BROKER_ID=${HOSTNAME##*-} && \
    export KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://${POD_IP}:9092 && \
    /etc/confluent/docker/run

Nevertheless, this patch seems to offer a cleaner solution.

PR #401, opened 2018-01-07
add kafka connect plugin path to the environment variable

evolv34:
Hi All,

All Connect plugins are not listing from localhost:8083/connector-plugins in kafka connect docker image. Its fixed in this PR. It would be great if somebody can take a look at this PR ?


perryao:
I had to do this myself. But not sure why it's necessary. FWIW, the file source and file sink are the only two that show up without this set.

evolv34:
This doesn’t list any other connectors like HDFS connectors or elastic search connectors that available out of the box in Confluent platform. So this fix would and list these connectors as well.

evolv34:
This fix would be helpful when developers refer to this yml file. Many developers use this as a point of reference while using containers.

johnseekins:
Agreed with above. The documentation implies that the built in plugins will be loaded by default. This isn't the case without @evolv34 's change.

cricket007:
Superseded by #464

evolv34:
Can it be merged ?

cricket007:
It was. See the other issue

PR #403, opened 2018-01-08
Detect ZOOKEEPER_SERVER_ID based on hostname in K8s 

### LINKED ISSUE #402: [cp-zookeeper] detect ZOOKEEPER_SERVER_ID in kubernetes
Currently `ZOOKEEPER_SERVER_ID` needs to be set otherwise container will fail to start.
Because `ZOOKEEPER_SERVER_ID` used as a template to fulfill zookeeper's `myid` file.
image 1
In Kubernetes environment, `myid` should be generated based on id of StatefulSet pod instance.

gAmUssA:
detects and generates zookeeper server id from the hostname.
If zookeeper deployed in StatefulSet, hostname will have a format `zookeeper-0`, `zookeeper-1`, etc.
The script detects the server ID based on the hostname.
StatefulSets are numbered from 0 so we have to always increment by 1

fixes #402

### DIFF
--- a/debian/zookeeper/include/etc/confluent/docker/configure
+++ b/debian/zookeeper/include/etc/confluent/docker/configure
@@ -25,8 +25,14 @@ dub ensure ZOOKEEPER_CLIENT_PORT
 dub path /etc/kafka/ writable
 
 # myid is required for clusters
+# Detect the ZOOKEEPER_SERVE_ID based on the hostname.
+# StatefulSets (kubernetes) are numbered from 0 so we have to always increment by 1
 if [[ -n "${ZOOKEEPER_SERVERS-}" ]]
 then
+  if [ -z "${ZOOKEEPER_SERVER_ID-unset}" ]; then
+    export ZOOKEEPER_SERVER_ID=$(hostname | awk -F'-' '{print $NF+1}')
+    echo "Detected Zookeeper ID $ZOOKEEPER_SERVER_ID"
+  fi
   dub ensure ZOOKEEPER_SERVER_ID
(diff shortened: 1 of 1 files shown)


gAmUssA:
[clabot:check]

mikesparr:
You could also implement similar changes to my PR #397 to the `zookeeper` image, instead editing the `run` script, allowing you to pass in args for environment variables.

This way you could do the following k8s manifest:
```yaml
args:
  - ZOOKEEPER_SERVER_ID=${HOSTNAME##*-}
```

by simply editing the `run` script adding the arg check
```
set -o nounset \
    -o errexit \
    -o verbose \
    -o xtrace


# Set environment values if they exist as arguments
if [ $# -ne 0 ]; then
  echo "===> Overriding env params with args ..."
  for var in "$@"
  do
    export "$var"
  done
fi

echo "===> ENV Variables ..."
env | sort

...
```

debian/zookeeper/include/etc/confluent/docker/configure
```diff
@@ -25,8 +25,14 @@ dub ensure ZOOKEEPER_CLIENT_PORT
 dub path /etc/kafka/ writable
 
 # myid is required for clusters
+# Detect the ZOOKEEPER_SERVE_ID based on the hostname.
+# StatefulSets (kubernetes) are numbered from 0 so we have to always increment by 1
 if [[ -n "${ZOOKEEPER_SERVERS-}" ]]
 then
+  if [ -z "${ZOOKEEPER_SERVER_ID-unset}" ]; then
```
cmccabe: Why are you using `if [ -z "${ZOOKEEPER_SERVER_ID-unset}"`?  It seems like that will never be true, since the `${FOO-bar}` construct gives the value "bar" if FOO is unset, right?

PR #424, opened 2018-02-07
Mesos TLS Kerberos Support

cyan-confluent:


debian/base/include/etc/confluent/docker/mesos-setup.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+if [ -z $SKIP_MESOS_AUTO_SETUP ]; then
```
maxzheng: How come u need $SKIP.. even though you are already checking MESOS_SANDBOX? As long as the image works when MESOS_SANDBOX is not set, then that seems safe.

debian/base/include/etc/confluent/docker/mesos-setup.sh
```diff
@@ -0,0 +1,28 @@
...
+        keytool -import \
+                -trustcacerts \
+                -alias root \
+                -file $MESOS_SANDBOX/.ssl/ca-bundle.crt \
+                -storepass changeit \
+                -keystore /tmp/kafka-truststore.jks -noprompt
+
+
+        if [ "$KERBEROS_ENABLED" = "true" ]; then
+            export KAFKAREST_OPTS="-Djava.security.auth.login.config=/tmp/client-jaas.conf -Djava.security.krb5.conf=/tmp/krb5.conf"
```
maxzheng: Who is providing /tmp/client-jaas.conf & /tmp/krb5.conf?

debian/base/include/etc/confluent/docker/mesos-setup.sh
cyan-confluent: yes it's safe for all non-users of mesos. But the idea here is that a small minority of mesos users may still prefer a manual configuration than an out of the box experience. I'm accommodating for that case.

debian/base/include/etc/confluent/docker/mesos-setup.sh
cyan-confluent: These two files are provided by secret configuration by:
This is on the marrthon/mesos side only, which is why you have to toggle manually in the mesos env:

dcos security secrets create -f /universe-workspace/confluent-aux-universe/config/client-jaas.conf jaas
dcos security secrets create -f /universe-workspace/confluent-aux-universe/config/krb5.conf krb5

PR #425, opened 2018-02-07
Mesos tls kerberos 4.0.0 post

cyan-confluent:


maxzheng (approved):
LGTM. Perhaps wait for @aayars to double check

cyan-confluent:
[clabot:check]

debian/base/include/etc/confluent/docker/mesos-setup.sh
```diff
@@ -0,0 +1,28 @@
...
+
+        keytool -import \
+                -trustcacerts \
+                -alias root \
+                -file $MESOS_SANDBOX/.ssl/ca-bundle.crt \
+                -storepass changeit \
+                -keystore /tmp/kafka-truststore.jks -noprompt
+
+
+        if [ "$KERBEROS_ENABLED" = "true" ]; then
```
cyan-confluent: Now that I think about it. I'm not sure if we really want this line. This line provides some convenience to the user, but if the user is already configuring secrets, he might as well be aware of configuring KAFKAREST_OPTS.

debian/base/include/etc/confluent/docker/mesos-setup.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+if [ -z $SKIP_MESOS_AUTO_SETUP ]; then
```
aayars: Can this check for a positive condition (like ENABLE_MESOS) instead?

cyan-confluent: Totally understand positives are preferred when coding. I'm making sure users who choose to skip this really knows what they are doing. They should be using the default setup unless they really really want to manually configure.

aayars (approved):
LGTM.

We're in the process of moving to new repos for all of these images, and will need to make this change there too. The new base won't be used for this release, though.

PR #428, opened 2018-02-12
DOCS-285: update config notes to reflect reality

cotedm:
drop config notes that no longer apply

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -51,13 +51,6 @@ Note: The Kafka Connect image is labeled as "Enterprise" simply because it conta
 Configuration Notes
 -------------------
 
-*  Docker for Mac
-
-	We do not recommend using these images with Docker for Mac at this time.  The primary reason is that Docker for Mac does not update the local /etc/hosts file with the hostnames of the deployed containers.   This makes it difficult to access the containerized cluster with client applications running directly on the Mac.  Additionally, the semantics of ``--net=host`` are not clear, so deploying containers with host networking on Docker for Mac is not reliable.  More details on these issues can be found at:
-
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -51,13 +51,6 @@ Note: The Kafka Connect image is labeled as "Enterprise" simply because it conta
 Configuration Notes
 -------------------
 
-*  Docker for Mac
-
-	We do not recommend using these images with Docker for Mac at this time.  The primary reason is that Docker for Mac does not update the local /etc/hosts file with the hostnames of the deployed containers.   This makes it difficult to access the containerized cluster with client applications running directly on the Mac.  Additionally, the semantics of ``--net=host`` are not clear, so deploying containers with host networking on Docker for Mac is not reliable.  More details on these issues can be found at:
```
ewencp: When did the `moby` thing get fixed? Is this actually targeted at the right branches? (And just for my own curiosity, do you know what we've done to fix it? I never got it running without all the extra `moby` host arguments, but I also haven't tried in quite a long while.)

joel-hamill:
i'll defer to eng on whether or not these apply.

docs/configuration.rst
cotedm: @ewencp great question and I don't know. cp-demo works without doing any `moby` gymnastics and I can't reproduce as I could when these images were first released. I've upgraded Docker for Mac a few times though:
Version 17.12.0-ce-mac49 (21995)

somewhere after 17.06 something changed it looks like? I don't see a relevant change from the images that would have made this happen. It's probably a good idea if you can confirm it's still broken for you before we merge this though.

PR #444, opened 2018-03-16
CC-1663: Added JMS connectors to Connect image

rhauch:
Also added test for the ActiveMQ connector. The Docker image includes JMS connector, ActiveMQ connector, and IBM MQ connector, but it’s very difficult to add messages to MQ (no command line utility) and so it’s going to be tough to write tests.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -97,7 +97,7 @@ Setup
          --amazonec2-region us-west-2 \
          --amazonec2-instance-type m4.large \
          --amazonec2-root-size 100 \
-         --amazonec2-ami ami-16b1a077 \
+         --amazonec2-ami ami-8f78c2f7 \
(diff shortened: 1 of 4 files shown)


docs/development.rst
```diff
@@ -97,7 +97,7 @@ Setup
          --amazonec2-region us-west-2 \
          --amazonec2-instance-type m4.large \
          --amazonec2-root-size 100 \
-         --amazonec2-ami ami-16b1a077 \
+         --amazonec2-ami ami-8f78c2f7 \
```
ewencp: why does this need to change?

rhauch: Because the old one doesn't work, and all of the required updated/dependencies don't install cleanly. I'll pull this out into a separate PR.

PR #449, opened 2018-03-21
CC-1456: Docker image for Replicator executable

kkonstantine:


wicknicks:
Looks good. A general comment. Given the number of mandatory and optional params this image can accept, adding a README which shows how to run this image would be super useful.

kkonstantine:
@wicknicks good point! I'm editing the rst files under `docs` currently. Given the pressing timeline, we might need a follow up PR for that. But if they are ready soon, I'll add the commit here.

gAmUssA:
Awesome, @kkonstantine !

debian/enterprise-replicator-executable/include/etc/confluent/docker/log4j.properties.template
```diff
@@ -0,0 +1,15 @@
+
+log4j.rootLogger={{ env["REPLICATOR_LOG4J_ROOT_LOGLEVEL"] | default('INFO') }}, stdout
+
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
+
+log4j.logger.org.reflections=ERROR
```
rhauch: What about these that are also in AK's log config for Connect?
```
log4j.logger.org.apache.zookeeper=ERROR
log4j.logger.org.I0Itec.zkclient=ERROR
```

docs/configuration.rst
```diff
@@ -472,3 +476,121 @@ Optional Settings
...
+  .. sourcecode:: bash
+
+    docker run -d \
+      --name=ReplicatorX \
+      --net=host \
+      -e REPLICATOR_LOG4J_ROOT_LOGLEVEL=DEBUG \
+      -v /mnt/replicator/config:/etc/replicator \
+      confluentinc/cp-enterprise-replicator-executable:4.1.0
+
+will start Replicator given that the directory ``/mnt/replicator/config``, that will be mounted under ``/etc/replicator`` on the Docker image, contains the required files ``consumer.properties``, ``producer.properties`` and the optional but often necessary file ``replication.properties``.
```
rhauch: What happens if the directories are not properly mounted or the required files are not available?

rhauch: Nit: maybe say "given that the _local_ directory" (without emphasis) to help users avoid mounting incorrectly.

debian/enterprise-replicator-executable/include/etc/confluent/docker/log4j.properties.template
kkonstantine: This is template that we is copied to replicator's log4j above. We need to suppress these warning because we embed the connect worker. Other than this template that will result to replicator's `log4j.properties` there's no other log4j properties file I think.

docs/configuration.rst
kkonstantine: I'd expect similar failures as when you don't set an environment variable. We'll explore such errors during testing and smoke testing of the images

kkonstantine:
Zk log4j suppressions added.

PR #465, opened 2018-04-20
DOCS-405 - Correct usage of "its" vs "it's" in cp-docker-images

mcqueary:
Fixes a few place where "it's" is mistakenly used instead of "its".


mcqueary:
[clabot:check]

PR #475, opened 2018-05-03
C3 configuration additions

coughman:


ewencp (commented):
@coughman is this something we still need? and if so, I think we'd need to make sure it also got ported to control-center-images since this repo has been deprecated and split up.

Also seems like maybe if we need to add more monitoring interceptor settings we should take them out of the `other_props` and use another `env_to_props` call that's specific to the `confluent.monitoring.interceptor.` prefix so we don't have to add settings individually -- at least at first glance it looks like the conversion is consistent across all the `CONTROL_CETNER_MONITORING_INTERCEPTOR_` settings.

PR #481, opened 2018-05-10
Just use kafkacat system package

aayars:


debian/kafkacat/Dockerfile
```diff
@@ -1,31 +1,18 @@
...
-    && git clone https://github.com/edenhill/kafkacat \
-    && cd kafkacat \
-    && git checkout tags/debian/$VERSION \
-    && ./bootstrap.sh \
-    && make install \
-    && cd .. && rm -rf kafkacat-debian-$VERSION \
-    && AUTO_ADDED_PACKAGES=`apt-mark showauto`  \
-    && apt-get remove --purge -y $BUILD_PACKAGES $AUTO_ADDED_PACKAGES \
-    \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y kafkacat \
```
maxzheng: Are we letting the OS apt repo to dictate the kafkacat version, or are we trying we don't care which version? 

Also, for `apt-get remove $AUDO_ADDED_PACKAGES`, that seems to be removing about 100MB of unnecessary stuff -- look at the size changes at https://hub.docker.com/r/confluentinc/cp-kafkacat/tags/ (except 3.2.2, which is big / perhaps it was built differently?). Perhaps just restore that / do the same for all other images as a separate change as it does seem to remove apt and other low level stuff?

maxzheng (approved):
If you are certain we don't care about the kafkacat version for each CP release, then sure.

debian/kafkacat/Dockerfile
aayars: The purge command seems to just obliterate the image, so I've left it out here.

I think it's okay, for kafkacat, to just pick up the current version of the system package.

edenhill: If system package means debian package it might be outdated and might depend on an even older librdkafka package.

So I suggest building the client from source instead.
This is what the upstream kafkacat Dockerfile does:
https://github.com/edenhill/kafkacat/blob/master/Dockerfile

aayars: Hey, that's a nice Dockerfile.

I feel like we should just build and tag the image out of the actual kafkacat repo.

PR #482, opened 2018-05-11
Build image for kafkacat directly out of actual repo

aayars:


Makefile
```diff
@@ -36,7 +36,12 @@ debian/base/include/etc/confluent/docker/docker-utils.jar:
 	&& cd -
 
 build-debian: debian/base/include/etc/confluent/docker/docker-utils.jar
-	for component in ${COMPONENTS} ; do \
+	pushd debian \
+	&& rm -rf kafkacat \
+	&& git clone https://github.com/edenhill/kafkacat \
+	&& sed -ie "s/runtimeDeps='/runtimeDeps='openssl libsasl2-modules-gssapi-mit krb5-user krb5-config /" kafkacat/Dockerfile \
```
jarekr: I guess `libssl1.1` and `libsasl2-2` are omitted intentionally?

jarekr (approved):

Makefile
aayars: Yeah -- it's not obvious, but the kafkacat Dockerfile already has them.

Makefile
```diff
@@ -36,7 +36,12 @@ debian/base/include/etc/confluent/docker/docker-utils.jar:
...
+	&& popd \
+	&& for component in ${COMPONENTS} ; do \
```
maxzheng: Might be cleaner to add another build target (e.g. kafkacat-Dockerfile) that this build target depends to do the checkout/Dockerfile mocking.

Also, that Dockerfile doesn't have these set: 

```
-ARG COMMIT_ID=unknown
-LABEL io.confluent.docker.git.id=$COMMIT_ID
-ARG BUILD_NUMBER=-1
-LABEL io.confluent.docker.build.number=$BUILD_NUMBER
-LABEL io.confluent.docker=true
```
And it's built using debian:stretch-slim instead of debian:stable. 

I like reusable, but not sure if that Dockerfile is too reusable for this as there are some differences. Having the Dockerfile in each repo is the right thing, but how do we ensure consistency for a packaged product (CP)?

aayars:
I'm going to try something less tenuous here, closing without merging.

PR #485, opened 2018-05-15
DOCS-462: updated cp-all-in-one docker compose with KSQL and monitori…

ybyzek:
…ng interceptors


ybyzek:
@joel-hamill got this working in his environment, and he gave the 👍 to merge on Zoom call

.gitignore
```diff
@@ -9,7 +9,6 @@ __pycache__
 *.srl
 *.req
 *.log
-*.jar
```
aayars: We absolutely can not support JARs checked into source control. Let's discuss this before propagating this change any further.

ybyzek: @aayars : ok, please back it out!  We'll work with @apurvam on a proper solution

PR #492, opened 2018-05-23
DEVX-97: adding REST proxy to cp-all-in-one

ybyzek:


edenhill:
kafka-rest is now a core team component

examples/cp-all-in-one/docker-compose.yml
```diff
@@ -122,3 +122,18 @@ services:
     environment:
       KSQL_CONFIG_DIR: "/etc/ksql"
       KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/etc/ksql/log4j-rolling.properties"
+
+  rest-proxy:
+    image: confluentinc/cp-kafka-rest:4.1.0
```
gAmUssA: 4.1.1 is the latest version

ijuma: The version issue affects all other products too. Not sure why the version wasn't updated when 4.1.1 was release.

ybyzek: Is it not the case that the `4.1.0-post` branch should use docker image 4.1.0, and the `4.1.1-post` branch should use docker image 4.1.1, etc?  (this question applies to all container images, not just the rest proxy).  And if yes, how does the image version get properly propagated branch to branch?  @aayars

ybyzek:
Ping @mageshn , @aayars

examples/cp-all-in-one/docker-compose.yml
```diff
@@ -91,3 +91,39 @@ services:
...
+      - $PWD/ksql-datagen-jars/ksql-examples-4.1.0.jar:/usr/share/java/ksql-cli/ksql-examples-4.1.0.jar
+      - $PWD/ksql-datagen-jars/kafka-connect-avro-converter-4.1.0.jar:/usr/share/java/ksql-cli/kafka-connect-avro-converter-4.1.0.jar
+      - $PWD/ksql-datagen-jars/generex-1.0.2.jar:/usr/share/java/ksql-cli/generex-1.0.2.jar
+      - $PWD/ksql-datagen-jars/automaton-1.11-8.jar:/usr/share/java/ksql-cli/automaton-1.11-8.jar
+    environment:
+      KSQL_CONFIG_DIR: "/etc/ksql"
+      KSQL_LOG4J_OPTS: "-Dlog4j.configuration=file:/etc/ksql/log4j-rolling.properties"
+
+  rest-proxy:
+    image: confluentinc/cp-kafka-rest:4.1.0
```
mageshn: any reason we want to hardcode the tags here in the examples?

examples/cp-all-in-one/docker-compose.yml
```diff
@@ -91,3 +91,39 @@ services:
...
+    depends_on:
+      - zookeeper
+      - broker
```
mageshn: perhaps include SR also as a dependency?

examples/cp-all-in-one/docker-compose.yml
ybyzek: Hmmm, I'm hesitant to always make it latest because variables change sometimes.  But for this case with REST proxy, I think it's ok.  I changed to `latest`

examples/cp-all-in-one/docker-compose.yml
ybyzek: Ok

PR #504, opened 2018-06-11
CC-2102 CP Connect Docker images suppress Reflections warnings

cyrusv:
Apache Kafka already suppresses these in a way we prefer
Resolves CC-2102 
https://confluentinc.atlassian.net/browse/CC-2102

### DIFF
--- a/debian/kafka-connect-base/include/etc/confluent/docker/log4j.properties.template
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/log4j.properties.template
@@ -7,6 +7,8 @@ log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
 
 {% set default_loggers = {
   'org.reflections': 'ERROR',
+  'org.apache.zookeeper': 'ERROR',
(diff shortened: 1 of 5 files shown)


debian/kafka-connect-base/include/etc/confluent/docker/log4j.properties.template
```diff
@@ -5,9 +5,16 @@ log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
 
+{% set default_loggers = {
+  'org.reflections': 'ERROR',
```
rhauch: Should this default also include the following?

```
'org.apache.zookeeper': 'ERROR',
'org.I0Itec.zkclientr': 'ERROR',
```

cyrusv: Fixed

PR #507, opened 2018-06-14
more generic mapping of env variables to properties

xvrl:
This should fix some of our env config mapping issues and avoid missing configs in the future.


ybyzek:
@xvrl , to a user, what is the impact of this change to those configuration parameters, e.g. `confluent.monitoring.interceptor.topic.replication`? Default value? Required/optional?

xvrl:
@ybyzek I don't think I'm changing any existing behavior. The additional excludes are only for env vars that we already define earlier in the file so we don't override them

PR #511, opened 2018-06-21
Properly run ksql-cli to avoid CPU hog

ybyzek:


examples/cp-all-in-one/docker-compose.yml
```diff
@@ -137,6 +137,8 @@ services:
       - broker
       - connect
       - ksql-server
+    entrypoint: /bin/sh
```
apurvam: Are you sure the `entrypoint` is strictly necessary? The cp-ksql-cli image already defines an entrypoint.

PR #516, opened 2018-06-27
 switch to multi-stage for kafkacat; bump to 1.3.1

aayars:
Installing from debs limited us to available versions in an external apt repo, and the old cleanup routine was nuking important system files. This just copies kafkacat into a mostly-pristine container, and installs runtime deps. Seems cleaner, and seems to work.

Pins the base image to stretch (use the same base as edenhill/kafkacat)

### DIFF
--- a/debian/kafkacat/Dockerfile
+++ b/debian/kafkacat/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stable
+FROM debian:jessie
 
 ARG COMMIT_ID=unknown
 LABEL io.confluent.docker.git.id=$COMMIT_ID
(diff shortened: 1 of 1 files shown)


debian/kafkacat/Dockerfile
```diff
@@ -1,14 +1,33 @@
-FROM debian:stable
+FROM debian:stretch-slim
+
+WORKDIR /build
+
+ENV VERSION=1.3.1-1
+
+ENV BUILD_PACKAGES="build-essential cmake python git curl zlib1g-dev libsasl2-dev libssl-dev"
+
+RUN echo "Building kafkacat ....." \
```
amberry: Because you are doing multistage, and therefore the layering won't matter, you could break this up to be more readable over multiple RUNs.

amberry (approved):

debian/kafkacat/Dockerfile
aayars: side by side they're about as readable, so i'll leave it for now.

PR #518, opened 2018-06-28
CC-1942. Kafka MQTT Docker image.

sansanichfb:


docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
 * Additional CLI options can be appended to the command as necessary. See :ref:`the KSQL CLI documentation <install_ksql-cli>` for the KSQL CLI options you can use.
 * The KSQL CLI cannot be configured via environment variables, unlike the KSQL Server image.
 
+-------------------------------
+Confluent Kafka MQTT Proxy
+-------------------------------
+
+For the Kafka MQTT Proxy image, use variables prefixed with ``KAFKA_MQTT_`` with an underscore (``_``) separating each word instead of periods. As an example, to set ``bootstrap.servers``, ``topic.regex.list`` you'd run the following:
```
kkonstantine: Kafka MQTT ~Proxy~

docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
...
+    docker run -d \
+      --name=cp-kafka-mqtt \
+      --net=host \
+      -e KAFKA_MQTT_BOOTSTRAP_SERVERS=PLAINTEXT://localhost:29092 \
+      -e KAFKA_MQTT_TOPIC_REGEX_LIST=mqtt:.* \
+      confluentinc/cp-kafka-mqtt:5.0.0
+
+Required Settings
+"""""""""""""""""
+The following settings must be passed to run the Kafka Connect Docker image:
```
kkonstantine: Kafka MQTT I guess you mean.

docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
...
+
+``KAFKA_MQTT_TOPIC_REGEX_LIST``
+
+  A comma-separated list of pairs of type '<kafka topic>:<regex>' that is used to map MQTT topics to Kafka topics.
+
+Optional Settings
+"""""""""""""""""
+All other settings for Kafka MQTT Proxy like security, producer overrides can be passed to the Docker images as environment variables. The names of these environment variables are derived by replacing ``.`` with ``_``, converting the resulting string to uppercase and prefixing it with ``KAFKA_MQTT_``. For example, if you need to set ``ssl.key.password``, the environment variable name would be ``KAFKA_MQTT_SSL_KEY_PASSWORD``.
```
kkonstantine: nit: I'd use singular since we are talking about a specific image: _to the Docker image_ ...

docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
...
+
+The image will then convert these environment variables to corresponding Kafka MQTT Proxy config variables.
```
kkonstantine: Kafka MQTT ~Proxy~

debian/kafka-mqtt/include/etc/confluent/docker/log4j.properties.template
```diff
@@ -0,0 +1,13 @@
+
+log4j.rootLogger={{ env["KAFKA_MQTT_LOG4J_ROOT_LOGLEVEL"] | default('INFO') }}, stdout
+
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
```
kkonstantine: This pr: https://github.com/confluentinc/newwave/pull/14 added the thread id and changed slightly the `log4j.properties`. I suggest we keep them identical for the optimal experience.

sansanichfb: Thanks, updated `ConversionPattern` according to the latest change.

docs/configuration.rst
sansanichfb: ✅

docs/configuration.rst
sansanichfb: ✅

docs/configuration.rst
sansanichfb: ✅

docs/configuration.rst
sansanichfb: ✅

docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
 * Additional CLI options can be appended to the command as necessary. See :ref:`the KSQL CLI documentation <install_ksql-cli>` for the KSQL CLI options you can use.
 * The KSQL CLI cannot be configured via environment variables, unlike the KSQL Server image.
 
+-------------------------------
+Confluent Kafka MQTT Proxy
```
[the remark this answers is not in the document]
sansanichfb: ✅

debian/kafka-mqtt/Dockerfile
```diff
@@ -0,0 +1,41 @@
+#
+# Copyright 2016 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
sansanichfb: It looks like this repos is licensed under Apache License 2.0, so we can keep headers.

debian/kafka-mqtt/Dockerfile
```diff
@@ -0,0 +1,41 @@
+#
+# Copyright 2016 Confluent Inc.
```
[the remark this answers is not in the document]
sansanichfb: Good catch, fixed.

docs/configuration.rst
```diff
@@ -664,3 +664,35 @@ Configuring the KSQL CLI
...
+Confluent Kafka MQTT
+-------------------------------
+
+For the Kafka MQTT image, use variables prefixed with ``KAFKA_MQTT_`` with an underscore (``_``) separating each word instead of periods. As an example, to set ``bootstrap.servers``, ``topic.regex.list`` you'd run the following:
+
+  .. sourcecode:: bash
+
+    docker run -d \
+      --name=cp-kafka-mqtt \
+      --net=host \
```
xvrl: examples should not use hot networking, since it does not work with docker for mac or windows

kkonstantine: @xvrl as you may confirm, this is not specific to this docker image. All of our example are using the above setting. Thus, a fix should be applied to all of them and should be the subject of a separate PR. The choice of @sansanichfb not to diverge here seems correct to me.

kkonstantine: Sure, I remember this fix in quick-starts. I also understand that this file was missed on that PR. We'll need a follow-up PR to fix it here too. Piggy-backing it in this PR can be misleading IMO. Good thing is, it's just docs so there's no code freeze for this change.

PR #575, opened 2018-08-23
CP-572: Restore ownership to root for AK / ZK

maxzheng:
The permissions got changed in https://github.com/confluentinc/kafka-packaging/pull/60/files#diff-6e14c4f3083a05273b2f12bdb8433776R19
And that does not work well for OpenShift and not relevant for running them in containers.

Testing: Built locally and works as expected

### DIFF
--- a/debian/enterprise-kafka/Dockerfile
+++ b/debian/enterprise-kafka/Dockerfile
@@ -36,9 +36,10 @@ RUN echo "===> installing confluent-support-metrics ..." \
     && echo "===> clean up ..."  \
     && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/* \
     \
-		&& echo "===> Setting up ${COMPONENT} dirs..." \
-   	&& mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
-   	&& chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets
+    && echo "===> Setting up ${COMPONENT} dirs..." \
+    && mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
+    && chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets \
+    && chown root:root /var/log/kafka /var/log/confluent /var/lib/kafka /var/lib/zookeeper
 
 VOLUME ["/var/lib/${COMPONENT}/data", "/etc/${COMPONENT}/secrets"]
 
--- a/debian/enterprise-kafka/Dockerfile
+++ b/debian/enterprise-kafka/Dockerfile
@@ -39,7 +39,7 @@ RUN echo "===> installing confluent-support-metrics ..." \
     && echo "===> Setting up ${COMPONENT} dirs..." \
     && mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
     && chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets \
-    && chown root:root /var/log/kafka /var/log/confluent /var/lib/kafka /var/lib/zookeeper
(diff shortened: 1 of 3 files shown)


debian/enterprise-kafka/Dockerfile
```diff
@@ -36,9 +36,10 @@ RUN echo "===> installing confluent-support-metrics ..." \
     && echo "===> clean up ..."  \
     && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/* \
     \
-		&& echo "===> Setting up ${COMPONENT} dirs..." \
-   	&& mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
-   	&& chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets
+    && echo "===> Setting up ${COMPONENT} dirs..." \
+    && mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
```
maxzheng: Converted tab to spaces. Change diff settings to hide whitelist changes. Same in kafka/Dockerfile.

mohnishbasha (approved):
LGTM

debian/enterprise-kafka/Dockerfile
```diff
@@ -36,9 +36,10 @@ RUN echo "===> installing confluent-support-metrics ..." \
...
+    && chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets \
+    && chown root:root /var/log/kafka /var/log/confluent /var/lib/kafka /var/lib/zookeeper
```
edenhill: should probably be recursive: `chown -R ..`

debian/kafka/Dockerfile
```diff
@@ -40,9 +40,10 @@ RUN echo "===> installing ${COMPONENT}..." \
     && echo "===> clean up ..."  \
     && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/* \
     \
-		&& echo "===> Setting up ${COMPONENT} dirs..." \
-   	&& mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
-   	&& chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets
+    && echo "===> Setting up ${COMPONENT} dirs..." \
+    && mkdir -p /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets\
+    && chmod -R ag+w /etc/${COMPONENT} /var/lib/${COMPONENT}/data /etc/${COMPONENT}/secrets \
+    && chown root:root /var/log/kafka /var/log/confluent /var/lib/kafka /var/lib/zookeeper
```
edenhill: make recursive

debian/zookeeper/Dockerfile
```diff
@@ -35,7 +35,8 @@ RUN echo "===> installing ${COMPONENT}..." \
     \
     && echo "===> Setting up ${COMPONENT} dirs" \
     && mkdir -p /var/lib/${COMPONENT}/data /var/lib/${COMPONENT}/log /etc/${COMPONENT}/secrets \
-    && chmod -R ag+w /etc/kafka /var/lib/${COMPONENT}/data /var/lib/${COMPONENT}/log /etc/${COMPONENT}/secrets
+    && chmod -R ag+w /etc/kafka /var/lib/${COMPONENT}/data /var/lib/${COMPONENT}/log /etc/${COMPONENT}/secrets \
+    && chown root:root /var/log/kafka /var/log/confluent /var/lib/kafka /var/lib/zookeeper
```
edenhill: dito

edenhill (changes requested):

debian/enterprise-kafka/Dockerfile
maxzheng: Note that your original change doesn't use recursive, so they are only changed for the top level and not necessary here.
There are subdirs still owned by root, but they are empty and also world writeable. 

I will update this to use recursive, but you might want to do the same in kafka-packaging if it affects systemd/security.

PR #576, opened 2018-08-23
Bump CP / AK versions in Makefile for 4.1.2

maxzheng:


mohnishbasha (approved):
LGTM, update final wiki as well. I will be looking into tbr to make bump for makefile.

PR #577, opened 2018-08-23
DOCS-735 - Set MySQL version to latest

joel-hamill:
Set MySQL download version to latest.


joel-hamill:
ping @rhauch for review

PR #578, opened 2018-08-27
Removed hardcoded versions from compose

mohnishbasha:
Removed hardcoded versions from compose in cp-all-in-one

### DIFF
--- a/examples/cp-all-in-one/docker-compose.yml
+++ b/examples/cp-all-in-one/docker-compose.yml
@@ -2,7 +2,7 @@
 version: '2'
 services:
   zookeeper:
-    image: confluentinc/cp-zookeeper
+    image: confluentinc/cp-zookeeper:5.0.0
     hostname: zookeeper
     container_name: zookeeper
     ports:
@@ -12,7 +12,7 @@ services:
       ZOOKEEPER_TICK_TIME: 2000
 
   broker:
-    image: confluentinc/cp-enterprise-kafka
+    image: confluentinc/cp-enterprise-kafka:5.0.0
     hostname: broker
     container_name: broker
     depends_on:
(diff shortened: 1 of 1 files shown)


examples/cp-all-in-one/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 version: '2'
 services:
   zookeeper:
-    image: confluentinc/cp-zookeeper:5.0.0-beta30
+    image: confluentinc/cp-zookeeper
```
aayars: ... running from latest is a worst practice. Why don't we just pin to the correct release tag?

aayars (changes requested):
Strongly object to this change. Can we bump versions properly with tooling instead?

examples/cp-all-in-one/docker-compose.yml
ybyzek: +1000 this is what I've been asking for in Slack.

mohnishbasha: @aayars So technically, the code within cp-all-in-one should point to latest release.
However, being in master makes it a bit ambiguous  since master holds -SNAPSHOT version.

I will pin it to 5.0.0, automating via tooling is a WIP.

ybyzek (approved):

aayars:
Master should be on 5.1.0.

mohnishbasha:
Yeva wanted this to be pinned to latest release.
5.1.0 is not release yet.

aayars:
Right now, master corresponds to 5.1.0.

Setting this should be a matter of course consistency across branches -- if someone needs 5.0.0, they should get it out of 5.0.0-post. mixing and matching versions in trunk isn't great.

mohnishbasha:
Ok, will do the following changes:
- Pin it to 5.1.0 in master. However, the compose files depends on 5.1.0.jar.
- Pin it to 5.0.0 in 5.0.0-post

ybyzek:
Actually my original request for `5.0.0-post` to pin to 5.0.0

mohnishbasha:
PR's are up:
https://github.com/confluentinc/cp-docker-images/pull/581
https://github.com/confluentinc/cp-docker-images/pull/580

PR #598, opened 2018-09-14
Add note about networking with link to article

rmoff:


rmoff:
@cricket007 good idea. I'll log a ticket internally for this.

PR #602, opened 2018-09-21
Add second listener and some text explaining its use

rmoff:
We need an example that shows people how to wire Kafka on Docker for both internal and external connections.

### DIFF
--- a/examples/kafka-single-node/docker-compose.yml
+++ b/examples/kafka-single-node/docker-compose.yml
@@ -3,23 +3,37 @@ version: '2'
 services:
   zookeeper:
     image: confluentinc/cp-zookeeper:latest
-    network_mode: host
     environment:
-      ZOOKEEPER_CLIENT_PORT: 32181
+      ZOOKEEPER_CLIENT_PORT: 2181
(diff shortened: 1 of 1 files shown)


ybyzek:
@rmoff : the enhancement to the `KAFKA_ADVERTISED_LISTENERS` parameter is great. With regard to the added comment, please consider (a) augmenting docs.confluent.io with this information and (b) removing the comment from the docker-compose file and instead adding a link to the docs?

rmoff:
@ybyzek (a) sure, I can suggest adding this (b) why remove it? It's not going to go out of date as such, and makes it more accessible for devs (and more likely they'll read it)

ybyzek:
My thinking is that if anything changes, you need to chase it down in duplicate places.  This is how things fall out of date.

In terms of getting devs to read it, would a link to docs from docker-compose file not get them there?

Or.... add a README?

rmoff:
I'm going to merge this in as-is, and track the docs improvement separately. Once made, I'll replace the comment with the docs URL.

rmoff:
Internal issue id DOCS-822

PR #606, opened 2018-10-02
CP-578: Add option to turn on trace and hide passwords by default.

maxzheng:
Trace will default to off as printing out each line could potentially
show the password in STDOUT. This will improve security by default, but
allow trace to be turned on if needed for debugging.

Testing: Verified locally by exporting the appropriate env vars

Note:
* After this is approved, the same will be done for new repos using common-docker, such as ksql.

### DIFF
--- a/debian/base/include/etc/confluent/docker/bash-config
+++ b/debian/base/include/etc/confluent/docker/bash-config
@@ -21,3 +21,8 @@ if [ "${TRACE:-}" == "true" ]; then
   set -o verbose \
       -o xtrace
 fi
+
+
+function show_env {
+    env | sort | grep -vP 'PASSWORD|JAAS_CONFIG'
+}
--- a/debian/base/include/etc/confluent/docker/bash-config
+++ b/debian/base/include/etc/confluent/docker/bash-config
@@ -1,6 +1,5 @@
-#!/usr/bin/env bash
 #
-# Copyright 2016 Confluent Inc.
+# Copyright 2018 Confluent Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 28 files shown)


debian/kafka/include/etc/confluent/docker/run
```diff
@@ -30,7 +26,7 @@ if [ $# -ne 0 ]; then
 fi
 
 echo "===> ENV Variables ..."
-env | sort
+env | sort | grep -vP 'PASSWORD|JAAS_CONFIG'
```
mohnishbasha: what if this list becomes long? May be use a constant string in bash-config so that all such env to be ignored can be listed at one place.

mohnishbasha (approved):
LGTM, left some comments.

debian/base/include/etc/confluent/docker/bash-config
```diff
@@ -0,0 +1,28 @@
...
+
+# Trace may expose passwords/credentials by printing them to stdout, so turn on with care.
+if [ "${TRACE:-}" == "true" ]; then
+  set -o verbose \
+      -o xtrace
+fi
+
+
+function show_env {
+    env | sort | grep -vP 'PASSWORD|JAAS_CONFIG'
```
mohnishbasha: This is much better.

debian/kafka/include/etc/confluent/docker/run
mcrocker-confluent: this fits in with what I was hoping to see. 

regarding mohnishbasha , its possible this list could grow . 

I also wonder about externalizing this to a common script like the bash_config that was done so customers/support would have an easy way to add env variables that are of particular concern to them.

PR #631, opened 2018-11-13
CC-2963: Kafka-MQTT docker image to support security on the readiness check.

avocader:


avocader:
@kkonstantine I've added docs, which describe configuration parameters for the image. Also, Docker compose example is included, which creates Kafka+ZK+MQTT Proxy cluster and configures MQTT Proxy to use an SSL-enabled producer.

kkonstantine:
@avocader sounds great. Also a good opportunity to finish and merge the relevant PR on the security-based quickstart once we get the chance, hopefully before `5.1`
Still LGTM

PR #639, opened 2018-11-29
DOCS-675 - Replace uses of docker machine

joel-hamill:
https://confluentinc.atlassian.net/browse/DOCS-675

## Staging Server

http://staging-docs.confluent.io.s3-website-us-west-2.amazonaws.com/PR/1415/5.0.0/installation/docker/docs/installation/replicator.html


docs/installation/automatic-data-balancing.rst
```diff
@@ -3,108 +3,36 @@
...
 
-This tutorial runs Confluent Auto Data Balancing (ADB) on Kafka, which allows you to shift data to create an even workload across your cluster.  By the end of this tutorial, you will have successfully run Confluent Auto Data Balancing CLI to rebalance data after adding and removing brokers.
+This tutorial runs :ref:`Confluent Auto Data Balancing (ADB) on Kafka <rebalancer>`, which allows you to shift data to create an even workload across your cluster.  By the end of this tutorial, you will have successfully run Confluent Auto Data Balancing CLI to rebalance data after adding and removing brokers.
 
-.. include:: includes/docker-tutorials.rst
-    :start-line: 2
+.. include:: includes/start-download.rst
 
-#. Start the services by using the example Docker Compose file.  Navigate to ``cp-docker-images/examples/enterprise-kafka``, where it is located:
+#.  Clone the Docker images repository, checkout the |release| branch, and navigate to the directory with the multi-datacenter example.
```
ybyzek: `navigate to the directory with the multi-datacenter example.` -> I don't think that's what you intended here for ADB?

docs/installation/automatic-data-balancing.rst
```diff
@@ -219,23 +141,17 @@ This tutorial runs Confluent Auto Data Balancing (ADB) on Kafka, which allows yo
 
    .. codewithvars:: bash
 
-    docker run \
-      --net=host \
-      --rm \
-      confluentinc/cp-enterprise-kafka:|release| \
-      bash -c "confluent-rebalancer status --zookeeper localhost:22181"
+    docker-compose exec broker \
+    bash -c "confluent-rebalancer status --zookeeper localhost:22181"
```
ybyzek: I think this should fail because the exec is on `broker` container, but the zookeeper connection is `localhost` (which is wrong).  It should be something like `--zookeeper <name of zookeeper container>:22181"`

docs/installation/automatic-data-balancing.rst
joel-hamill: copy/paste error

docs/includes/create-secret.rst
```diff
@@ -0,0 +1,42 @@
...
+    Linux Users
+        Run the following commands on the Linux host.
+
+        .. codewithvars:: bash
+
+            export ETH0_IP=$(ifconfig eth0 | grep 'inet addr:' | cut -d: -f2 | awk '{ print $1}')
+
+            echo ${ETH0_IP} quickstart.confluent.io >> /etc/hosts
+
+    MacOS and Windows Users
```
londoncalling, line 34: @ybyzek , @joel-hamill I am updating the Docs per https://confluentinc.atlassian.net/browse/DOCS-3656, which references this PR, to remove the prerequisite of `docker-machine`. But it looks like you missed one reference to `docker-machine` here. If we are no longer requiring `docker-machine`, we shouldn't be telling them to `ssh` into the container using `docker-machine` right? 

Does this `include` also need to be further updated to remove `docker-machine` reference?

PR #647, opened 2018-12-12
Add zookeeper-connect in docker-compose example for rest proxy

stanislavkozlovski:
It could not establish a zookeeper connection on the default `localhost:2181`. We need to pass the ZK address


stanislavkozlovski:
the REST proxy has always needed a zookeeper connection for the v1 consumers. I imagine that this example never worked with the v1 API consumers.

stanislavkozlovski:
@ybyzek good thinking. I'll cut a new branch from 5.0.1

PR #648, opened 2018-12-12
Add zookeeper-connect in docker-compose example for rest proxy

stanislavkozlovski:


gAmUssA:
@stanislavkozlovski out of curiosity, why bother with zookeeper connect?

stanislavkozlovski:
@gAmUssA 

the REST proxy has always needed a zookeeper connection for the v1 consumers. I tested the v1 api and saw that it didn't work because it tried to connect to the default `localhost:2181`
I imagine that this example has never worked with the v1 API consumers.

gwenshap:
@stanislavkozlovski should we merge or close?

andrewegel:
Next time please pint-merge these changes up.

PR #676, opened 2019-01-23
DEVX-662: Monitoring Replicator in Control Center

ybyzek:
Focus of PR review should be on https://github.com/confluentinc/cp-docker-images/blob/DEVX-662/examples/multi-datacenter/README.md#monitor-replicator-performance

Note: prereq PR: https://github.com/confluentinc/examples/pull/107

PR #682, opened 2019-02-07
Light changes on punctuation and hyphen usage

vsyu:
A few comments in addition to the suggested changes: 1) Should "Consumer Timestamps Interceptor" be capitalized (similar to https://docs.confluent.io/4.0.1/control-center/docs/installation/clients.html)? Features and names of a particular component should be in caps, but we should avoid capitalization where it isn't necessary so just wanted to confirm. 2) From the outside it feels like there should be some article (e.g., the, your) before DC1 in "run the consumer to connect to DC1 Kafka cluster," and same for "connect to DC2 Kafka cluster."

### DIFF
--- a/examples/multi-datacenter/README.md
+++ b/examples/multi-datacenter/README.md
@@ -1,9 +1,9 @@
 # Overview
 
-This example deploys an active-active multi-datacenter design, with two instances of Confluent Replicator copying data bi-directionally between the datacenters.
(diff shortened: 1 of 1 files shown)


ybyzek:
I ported these comments into the existing PR https://github.com/confluentinc/cp-docker-images/pull/676

PR #683, opened 2019-02-07
Small punctuation changes

vsyu:
I'm not sure if "consumer lag" should be capitalized. It looks like something we monitor but in it of itself isn't a capability or formal name for it. Also, regarding the sentence, "...`replicator-dc1-to-dc2-topic2` to view Replicator's consumer log," should it actually be "lag"?

### DIFF
--- a/examples/multi-datacenter/README.md
+++ b/examples/multi-datacenter/README.md
@@ -66,23 +66,23 @@ Stop all services:
 
 Monitoring Replicator is important to:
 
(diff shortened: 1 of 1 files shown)


ybyzek:
@vsyu sorry, this is my mistake, I saw this PR only now after I merged DEVX-662.  So I created a new PR and ported all your suggestions there: https://github.com/confluentinc/cp-docker-images/pull/691

PR #687, opened 2019-02-11
Move docker-utils to root folder 

maxzheng:
This will allow our release script to properly set the version using mvn. This was done in 5.1.x, so we are doing the same thing for older branches to get a consistent release experience.


maxzheng:
Testing: Ran make commands and got the same result as before

jarekr (approved):

yangxi (approved):

PR #704, opened 2019-03-01
Always require truststore when using TLS.

rkettelerij:
A truststore isn't just needed for TLS client auth, you'll always need a truststore for TLS to work if you're using self-signed certs. E.g: without the truststore settings interbroker communication over TLS won't work. This results in `PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target` errors due to missing CA certs. Hence you'll need a truststore.

UPDATE: looks like #611 has the same request.


rkettelerij:
[clabot:check]

rkettelerij:
[clabot:check]

rkettelerij:
Any update on this?

ChristianCiach:
I can't believe this is still open. Funny enough, this is a regression. Like five years ago, the code looked exactly like you are proposing in this PR.

Fortunately, the issue can easily be workarounded by using the variable ` KAFKA_SSL_TRUSTSTORE_LOCATION` directly instead of `KAFKA_SSL_TRUSTSTORE_FILENAME`.

PR #715, opened 2019-03-26
Remove *BASIC_AUTH_USER_INFO fields from show_env

ChrisStuntz:
We don't want to log any sensitive data on startup, so I would like to exclude all of the BASIC_AUTH_USER_INFO fields from the `show_env` function. This will make sure that no sensitive data gets logged and sent to any shared logging systems.


ChrisStuntz:
[clabot:check]

mnaseef-gloo:
Bump - this is a problem for us as well, and we would like to see this pull request approved.

PR #716, opened 2019-03-26
Debian has removed mirrors for jessie, switch to archives

jarekr:
We will migrate from jessie to stretch most likely.

https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html

### DIFF
--- a/debian/base/Dockerfile
+++ b/debian/base/Dockerfile
@@ -63,6 +63,9 @@ ENV ZULU_OPENJDK_VERSION="8=8.30.0.1"
 ENV LANG="C.UTF-8"
 
 RUN echo "===> Updating debian ....." \
+    # TODO debian jessie has been deprecated and is only ex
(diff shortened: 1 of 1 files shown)


debian/base/Dockerfile
```diff
@@ -63,6 +63,9 @@ ENV ZULU_OPENJDK_VERSION="8=8.30.0.1"
 ENV LANG="C.UTF-8"
 
 RUN echo "===> Updating debian ....." \
+    # TODO debian jessie has been deprecated and is only ex
+    && echo "deb http://archive.debian.org/debian/ jessie main" > /etc/apt/sources.list \
+    && echo "deb http://security.debian.org jessie/updates main" >> /etc/apt/sources.list \
```
maxzheng: Do we need to do this in older branches?

maxzheng (approved):

PR #717, opened 2019-03-26
Debian has removed mirrors for jessie, switch to archives

jarekr:
We will migrate from jessie to stretch most likely.

https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html


maxzheng (approved):

scottball1968:
When could we expect a confluent image based on debian stretch, or newer version?

PR #720, opened 2019-04-02
secure defaults

nineinchnick:
* fetch gpg keys from secure endpoint using full fingerprint
* use secure Confluent repo by default


nineinchnick:
[clabot:check]

maxzheng (approved):
Nice. Thanks for contributing!

PR #722, opened 2019-04-03
DEVX-791: Use port 9092 instead of 29092 for connecting to broker fro…

ybyzek:
…m host

Most users are used to connecting to a broker on port 9092 from their host, so this PR will help prevent user misconfiguration (e.g. https://github.com/confluentinc/examples/issues/142)


examples/cp-all-in-one/docker-compose.yml
```diff
@@ -18,17 +18,17 @@ services:
     depends_on:
       - zookeeper
     ports:
-      - "9092:9092"
       - "29092:29092"
```
rmoff: Nit: this doesn't need exposing if it's only being used within the Docker network. It's only 9092 that needs exposing.

PR #725, opened 2019-04-09
MINOR: Depend on checkstyle version from common pom

ewencp:
The common repo is now on a newer version and this is easier to manage centrally in one place unless we have a specific need to override in specific repos.


ybyzek (approved):
LGTM.  Must have been a copy/paste from elsewhere.

PR #733, opened 2019-05-08
fix parameter names

xli1996:
https://confluentinc.atlassian.net/browse/ST-1584

>In the launch script, the variable KAFKA_REST_JMX_PORT is being set. However, within /usr/bin/kafka-rest-run-class inside the container, that variable is never being used. Instead, the variable is KAFKAREST_JMX_OPTS. Because of this mismatch in variable names, the RMI and JMX Hostname parameters are never set.

There is a name issue between `cp-docker-images` and `kafka-rest`. After considering the compatbility, I decide to change `cp-docker-images`. From `KAFKA_REST_JMX_OPTS ` and `KAFKA_REST_JMX_PORT ` to `KAFKAREST_JMX_OPTS ` and `KAFKAREST_JMX_PORT `.

Since we still support 3.3.x, so I made the changes from 3.3.x.

In future, we may still need to consider the name conventions.

### DIFF
--- a/debian/kafka-rest/include/etc/confluent/docker/configure
+++ b/debian/kafka-rest/include/etc/confluent/docker/configure
@@ -30,9 +30,9 @@ then
   exit 1
 fi
 
-if [[ -n "${KAFKA_REST_JMX_OPTS-}" ]]
+if [[ -n "${KAFKAREST_JMX_OPTS-}" ]]
 then
-  if [[ ! $KAFKA_REST_JMX_OPTS == *"com.sun.management.jmxremote.rmi.port"*  ]]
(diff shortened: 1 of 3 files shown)


mohnishbasha (approved):
LGTM, however will be nice to have someone from kafka rest give a +1 as well.

gwenshap:
In https://confluent.slack.com/archives/C60B5SJ1E/p1557341424093100 We said:
we need to go with KAFKA_REST_JMX_OPTS

But this PR does the opposite. What changed?

xli1996:
> In https://confluent.slack.com/archives/C60B5SJ1E/p1557341424093100 We said:
> we need to go with KAFKA_REST_JMX_OPTS
> 
> But this PR does the opposite. What changed?

Since I do not know a lot about `kafka-rest`, I think it might be better to follow what `kafka-rest` is already used incase I break any thing.

PR #738, opened 2019-05-31
fix typo in the error message regarding KAFKA_JMX_OPTS

DaveSlinn:


DaveSlinn:
I didn't know about the different branch names and their purposes, so no, I had no intentions one way or the other for this change. I viewed it as just a typo, and a minor one at that, that could be fixed at any time.

ewencp:
Cool, adjusted and merging. Thanks for the contribution!

PR #751, opened 2019-06-30
switch default scala version to 2.12

jarekr:
use 2.12 scala artifacts, these should be used by default and not 2.11

confluent-server does not have it's scala version in the artifact name anymore


ijuma (approved):
LGTM cc @omkreddy

sdandu-gh (approved):
LGTM

PR #753, opened 2019-07-09
CC-5432: Add Confluent security plugins to Connect Docker image

C0urante:
[Jira](https://confluentinc.atlassian.net/browse/CC-5432)

We need to include the `confluent-security` package in our Docker image for Connect in order to support RBAC.

### DIFF
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -35,6 +35,8 @@ RUN echo "===> Installing Schema Registry (for Avro jars) ..." \
         confluent-schema-registry=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Controlcenter for monitoring interceptors ..."\
     && apt-get install -y confluent-control-center=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
+    && echo "===> Installing Confluent security plugins ..."\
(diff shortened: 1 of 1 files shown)


debian/kafka-connect-base/Dockerfile
```diff
@@ -35,6 +35,8 @@ RUN echo "===> Installing Schema Registry (for Avro jars) ..." \
         confluent-schema-registry=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Controlcenter for monitoring interceptors ..."\
     && apt-get install -y confluent-control-center=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
+    && echo "===> Installing Confluent security plugins ..."\
+    && apt-get install -y confluent-security=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
```
wicknicks: do we need to do this in `debian/kafka-mqtt/` as well?

C0urante: I don't _think_ so but I'm not 100%. We already include the `kafka-client-plugins-5.3.0-ce.jar` artifact (and its dependencies) in the MQTT proxy image since it's [listed as a dependency in the POM](https://github.com/confluentinc/newwave/blob/12e8f749a12c8e69e67b278bf335a409c1bd7487/kafka-mqtt/pom.xml#L36-L41), so I believe we have everything we need to run the proxy against an RBAC-enabled broker.

I'll make sure to manually verify this, but for now I think we can merge safely without including these changes in the MQTT proxy as well.

PR #757, opened 2019-07-12
CC-5454, CC-5456: Add new cp-server-connect-base and cp-server-connect images

C0urante:
[CC-5454](https://confluentinc.atlassian.net/browse/CC-5454) covers how we plan to add two new enterprise Connect images that derive from `cp-server` instead of `cp-kafka`, and revert the recent changes for CC-5432 which added the `confluent-security` package to the `cp-kafka-connect-base` image.
[CC-5456](https://confluentinc.atlassian.net/browse/CC-5456) covers how we should alter our Replicator images to derive from `cp-server` instead of `cp-kafka` as well.

The changes here address both issues. Two new images, `cp-server-connect-base`, and `cp-server-connect`, are added; they mirror the existing `cp-kafka-connect-base` and `cp-kafka-connect` images and the only difference is that `cp-server-connect-base` inherits from `cp-server` instead of `cp-kafka`, `cp-server-connect` inherits from `cp-server-connect-base` instead of `cp-kafka-connect-base`, and `cp-server-connect-base` includes the `confluent-security` package. The Replicator image is altered to inherit from this new `cp-server-connect-base` image instead of the `cp-kafka-connect-base` image. Because the Replicator Executable image inherits from the regular Replicator image, it too is transitively affected, though it isn't explicitly changed in this PR.

**NOTE FOR REVIEWERS**: Don't be fooled by the size of this PR; a recursive diff between `debian/kafka-connect-base` and `debian/server-connect-base`, and between `debian/kafka-connect` and `debian/server-connect`, clearly demonstrates that the new images are near-identical copies of existing ones and the actual changes are very small. Because this PR is a release blocker, please refrain from reviewing the unchanged parts of the new images that come directly from the existing images unless there is either a bug in both or some new difference between the two is warranted.

Both new images have been tested manually and verified. Automated Python tests have not yet been added for them; if those are deemed important enough to be release blockers, I can add them to this PR as well.

### DIFF
--- a/debian/enterprise-replicator/Dockerfile
+++ b/debian/enterprise-replicator/Dockerfile
@@ -15,7 +15,7 @@
 
 # Builds a docker image for Replicator
 
-FROM confluentinc/cp-kafka-connect-base
+FROM confluentinc/cp-server-connect-base
 
 MAINTAINER partner-support@confluent.io
 LABEL io.confluent.docker=true
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -35,8 +35,6 @@ RUN echo "===> Installing Schema Registry (for Avro jars) ..." \
         confluent-schema-registry=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Controlcenter for monitoring interceptors ..."\
     && apt-get install -y confluent-control-center=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
-    && echo "===> Installing Confluent security plugins ..."\
-    && apt-get install -y confluent-security=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Confluent Hub client ..."\
     && apt-get install -y confluent-hub-client=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Cleaning up ..."  \
--- a/debian/server-connect-base/Dockerfile
+++ b/debian/server-connect-base/Dockerfile
@@ -1,0 +1,52 @@
+#
+# Copyright 2016 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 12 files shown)


debian/enterprise-replicator/Dockerfile
```diff
@@ -15,7 +15,7 @@
 
 # Builds a docker image for Replicator
 
-FROM confluentinc/cp-kafka-connect-base
+FROM confluentinc/cp-server-connect-base
```
stan-confluent: do we need to have both versions for replicator, one inheriting from cp-kafka and another from cp-server?

C0urante: Nope, Replicator is already an enterprise product so there's no need to have an additional image that inherits from our non-enterprise image.

C0urante:
After offline discussion with @sdanduConf we've decided to implement tests in a follow-up PR.

sdanduConf:
> Changes look good to me. As for tests, it'd be good to know where / when they are actually run to determine if they should block this (fyi, we added the cp-server image without new tests, but not sure that precedent is the desired state)

Though not really the desired state, i think we should be moving this forward without this particular new test.  I am still reviewing, but don't see these tests actually being run in any consistent manner. Separately, we will be explicitly doing testing (@C0urante and @stan-confluent ) on these images as soon as they are ready.  I will be filing follow-up JIRAs for 1) making sure tests get built for these 2 images 2) making sure we get one for cp-server 3) making sure current tests are actually working 4) getting those running regularly (either as part of overlay job or elsewhere).  Don't think it makes sense to block PR on that, but should happen as a short-term follow-up item.

debian/server-connect-base/Dockerfile
```diff
@@ -0,0 +1,52 @@
+#
+# Copyright 2016 Confluent Inc.
```
rhauch: Nit:
```suggestion
# Copyright 2019 Confluent Inc.
```

Note this may apply to all of the files being added in this PR.

debian/kafka-connect-base/Dockerfile
```diff
@@ -35,8 +35,6 @@ RUN echo "===> Installing Schema Registry (for Avro jars) ..." \
         confluent-schema-registry=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Controlcenter for monitoring interceptors ..."\
     && apt-get install -y confluent-control-center=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
-    && echo "===> Installing Confluent security plugins ..."\
-    && apt-get install -y confluent-security=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
     && echo "===> Installing Confluent Hub client ..."\
```
wicknicks: q: are there no consumers of this image that depend on these libraries?

C0urante: We just added those lines in [a few days ago](https://github.com/confluentinc/cp-docker-images/pull/753), so I certainly hope not.

C0urante:
Thanks for the reviews, guys!

PR #759, opened 2019-07-16
Remove mandatory internal converters configuration (deprecated AK2.0)

rmoff:
Fixed #229

PR #761, opened 2019-07-16
Remove mandatory internal converters configuration (deprecated AK2.0)

### LINKED ISSUE #229: cp-kafka-connect - Make internal converter env vars optional
`CONNECT_INTERNAL_KEY_CONVERTER` and `CONNECT_INTERNAL_VALUE_CONVERTER` environment variables need to be set when using the docker image. [link](https://github.com/confluentinc/cp-docker-images/blob/3.2.0-post/debian/kafka-connect-base/include/etc/confluent/docker/configure#L29-L30)
However considering that these should never be changed from the [default `JsonConverter` class](https://github.com/apache/kafka/blob/trunk/config/connect-distributed.properties#L35), I propose that these environment variables be defaulted to JsonConverter class and be optional. [discussion](https://confluentcommunity.slack.com/archives/connect/p1488991284001084)

rmoff:
(replaces #759)

Fixed #229

### DIFF
--- a/debian/kafka-connect-base/include/etc/confluent/docker/configure
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/configure
@@ -26,8 +26,6 @@ dub ensure CONNECT_OFFSET_STORAGE_TOPIC
 dub ensure CONNECT_STATUS_STORAGE_TOPIC
 dub ensure CONNECT_KEY_CONVERTER
 dub ensure CONNECT_VALUE_CONVERTER
-dub ensure CONNECT_INTERNAL_KEY_CONVERTER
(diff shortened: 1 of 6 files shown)


ybyzek:
@rmoff I think this PR should have been committed to 5.0.0-post.  Because this PR went into 5.0.x, it didn't make it into any of the existing 5.0 releases like 5.0.0, 5.0.1, etc.

cc: @rspurgeon

ybyzek:
Actually, I take that back.  I don't think it could go into 5.0.0-post because that image already went out.  (I got confused with our world of examples and docs where we can commit to post branches)

PR #766, opened 2019-07-24
Add Dockerfile for kafka-connect-datagen

rmoff:


debian/kafka-connect-datagen/Dockerfile
```diff
@@ -0,0 +1,28 @@
+#
+# Copyright 2016 Confluent Inc.
```
ewencp: nit: date should be updated here.

debian/kafka-connect-datagen/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Builds a docker image for Kafka Connect
+
+FROM confluentinc/cp-kafka-connect-base
+
+MAINTAINER devx@confluent.io
```
ewencp: I think we probably just want to be consistent with maintainership across the board (unless we really do dockerfile-specific team targeting), though I'm not sure why everything else is pointed at `partner-support@confluent.io`...

debian/kafka-connect-datagen/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+LABEL io.confluent.docker=true
+ARG COMMIT_ID=unknown
+LABEL io.confluent.docker.git.id=$COMMIT_ID
+ARG BUILD_NUMBER=-1
+LABEL io.confluent.docker.build.number=$BUILD_NUMBER
+
+RUN echo "===> Installing Datagen Source Connector ..."
+RUN confluent-hub install confluentinc/kafka-connect-datagen:latest --no-prompt
```
ewencp: are we concerned about compatibility? as this gets maintained on older release branches, `latest` may no longer work with older connect versions (and in general just results in builds that aren't reproducible).

I assume the issue here is that we can't use the same approach re: versions as elsewhere since kafka-connect-datagen is versioned independently? If that's the case, we should probably introduce an `ENV` that sets it explicitly and can be put in place of `latest`.

debian/kafka-connect-datagen/Dockerfile
rmoff: Fixed

debian/kafka-connect-datagen/Dockerfile
rmoff: Reverted to partner-support

debian/kafka-connect-datagen/Dockerfile
rmoff: Added ARG and set to current latest version `0.1.2`

debian/kafka-connect-datagen/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+FROM confluentinc/cp-kafka-connect-base
+
+MAINTAINER partner-support@confluent.io
+LABEL io.confluent.docker=true
+ARG COMMIT_ID=unknown
+LABEL io.confluent.docker.git.id=$COMMIT_ID
+ARG BUILD_NUMBER=-1
+LABEL io.confluent.docker.build.number=$BUILD_NUMBER
+ARG DATAGEN_VERSION=0.1.2
```
ybyzek: Why 0.1.2?  The latest tag is `v0.1.3`

rmoff: I was going on what's listed on Confluent Hub, which is `0.1.2`

See @ewencp 's comments above re. `latest` https://github.com/confluentinc/cp-docker-images/pull/766#discussion_r307085720

debian/kafka-connect-datagen/Dockerfile
ybyzek: There is no good reason why kafka-connect-datagen isn't versioned with the rest of CP.  Shoul we perhaps change that first, and then we can use the same approach?

rmoff: ¯\_(ツ)_/¯ I don't mind either way :)

PR #742, opened 2019-06-14
Add support for Docker's healthcheck feature

rmoff:
Healthcheck looks at Kafka Connect REST endpoint and declares Kafka Connect 'heathly' if it gets HTTP 200 from it. 

```
$ docker ps
CONTAINER ID        IMAGE                                    COMMAND                   CREATED             STATUS                   PORTS                              NAMES
8138c423eeb5        confluentinc/cp-kafka-connect:5.2.1      "bash -c 'echo \"Inst…"   2 minutes ago       Up 2 minutes (healthy)   0.0.0.0:8083->8083/tcp, 9092/tcp   kafka-connect
```

PR #769, opened 2019-07-24
Add healthcheck for Kafka Connect

rmoff:
This adds support for the Healthcheck functionality in Docker to Kafka Connect, enabling more accurate orchestration and management of containers. 

It works with both HTTP and HTTPS. 

There is a five-second polling interval, which is therefore the maximum that this would add to any dependents in starting. 

(supersedes https://github.com/confluentinc/cp-docker-images/pull/742)

### No SSL

Using [zero-to-hero](https://github.com/confluentinc/demo-scene/tree/master/kafka-connect-zero-to-hero), mounted the healthcheck script as an external file, and added `healthcheck` to the Docker Compose: 

    healthcheck:
      timeout: 10s
      # 60 * 5s retries = five minutes
      interval: 5s
      retries: 120
      test: /etc/confluent/docker/healthcheck.sh
    volumes:
      - /Users/Robin/git/cp-docker-images/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh:/etc/confluent/docker/healthcheck.sh

Ran `docker-compose up -d`. 

Monitor healthcheck status: 

    $ docker inspect kafka-connect-01|jq '.[].State.Health'
    {
      "Status": "healthy",
      "FailingStreak": 0,
      "Log": [
        {
          "Start": "2019-07-24T10:57:11.1432877Z",
          "End": "2019-07-24T10:57:13.3144807Z",
          "ExitCode": 1,
          "Output": "Wed Jul 24 10:57:12 UTC 2019 \tKafka Connect HTTP state:  000  (waiting for 200)\n"
        },
        {
          "Start": "2019-07-24T10:57:18.3804919Z",
          "End": "2019-07-24T10:57:20.8000458Z",
          "ExitCode": 0,
          "Output": "Woohoo! Kafka Connect is up!\n"
        },

Container shows as `healthy`: 

    $ docker ps
    CONTAINER ID        IMAGE                                                 COMMAND                   CREATED             STATUS                   PORTS                                                      NAMES
    97bb1d396845        confluentinc/cp-enterprise-control-center:5.3.0       "bash -c 'echo \"Wait…"   4 minutes ago       Up 4 minutes             0.0.0.0:9021->9021/tcp                                     control-center
    04585a48c8e8        confluentinc/cp-ksql-cli:5.3.0                        "/bin/sh"                 4 minutes ago       Up 4 minutes                                                                        ksql-cli
    f7766e11d4d4        confluentinc/ksql-examples:5.1.2                      "bash -c 'echo Waiti…"    4 minutes ago       Up 4 minutes                                                                        datagen-ratings
    71f5a677268b        confluentinc/cp-kafka-connect:5.3.0                   "bash -c 'echo \"Inst…"   4 minutes ago       Up 4 minutes (healthy)   0.0.0.0:8083->8083/tcp, 9092/tcp                           kafka-connect-01
    […]

### With SSL

Using [cp-demo](https://github.com/confluentinc/cp-demo/), mounted the healthcheck script as an external file, and added `healthcheck` paragraph to the Docker Compose: 

    healthcheck:
      timeout: 10s
      # 60 * 5s retries = five minutes
      interval: 5s
      retries: 120
      test: /etc/confluent/docker/healthcheck.sh
    volumes:
      - /Users/Robin/git/cp-docker-images/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh:/etc/confluent/docker/healthcheck.sh
      […]

Ran `docker-compose up -d`. 

Monitor healthcheck status: 

    $ docker inspect connect|jq '.[].State.Health'
    {
      "Status": "healthy",
      "FailingStreak": 0,
      "Log": [
        {
          "Start": "2019-07-24T10:43:29.5857643Z",
          "End": "2019-07-24T10:43:38.5754631Z",
          "ExitCode": 1,
          "Output": "Wed Jul 24 10:43:38 UTC 2019 \tKafka Connect with SSL listener HTTP state:  000  (waiting for 200)\n"
        },
        {
          "Start": "2019-07-24T10:43:43.5995865Z",
          "End": "2019-07-24T10:43:45.8553504Z",
          "ExitCode": 0,
          "Output": "Woohoo! Kafka Connect with SSL listener is up!\n"
        },

Container shows as `healthy`: 

    $ docker ps
    CONTAINER ID        IMAGE                                                 COMMAND                  CREATED             STATUS                   PORTS                                                        NAMES
    03cd73bc6ff5        docker.elastic.co/kibana/kibana:5.5.2                 "/bin/sh -c /usr/loc…"   3 minutes ago       Up 3 minutes             0.0.0.0:5601->5601/tcp                                       kibana
    14545dea8585        confluentinc/cp-ksql-cli:5.3.0                        "/bin/sh"                3 minutes ago       Up 3 minutes                                                                          ksql-cli
    5d920ec3d2a3        docker.elastic.co/elasticsearch/elasticsearch:5.6.0   "/bin/bash bin/es-do…"   3 minutes ago       Up 3 minutes             0.0.0.0:9200->9200/tcp, 0.0.0.0:9300->9300/tcp               elasticsearch
    7c2a6e00ca81        confluentinc/cp-enterprise-control-center:5.3.0       "/etc/confluent/dock…"   3 minutes ago       Up 3 minutes             0.0.0.0:9021-9022->9021-9022/tcp                             control-center
    e0b56a3c343e        confluentinc/cp-ksql-server:5.3.0                     "/etc/confluent/dock…"   3 minutes ago       Up 3 minutes             0.0.0.0:8088->8088/tcp                                       ksql-server
    d1f5ad6e1d30        confluentinc/cp-kafka-connect:5.3.0                   "/etc/confluent/dock…"   3 minutes ago       Up 3 minutes (healthy)   0.0.0.0:8083->8083/tcp, 9092/tcp                             connect
    8265c8322bd0        confluentinc/cp-kafka-rest:5.3.0                      "/etc/confluent/dock…"   39 minutes ago      Up 37 minutes            8082/tcp, 0.0.0.0:8086->8086/tcp                             restproxy
    […]

### DIFF
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -54,5 +54,5 @@ CMD ["/etc/confluent/docker/run"]
 # Start-up period : 2 minutes (during which failures are not counted as failures)
 # Retry period    : 8 minutes (after which container is deemed unhealthy)
 # All settings can be overriden at run-time in Docker/Docker Compose. 
-HEALTHCHECK --start-period 120s --interval=5s --timeout=10s --retries=96 \
+HEALTHCHECK --start-period=120s --interval=5s --timeout=10s --retries=96 \
 	CMD /etc/confluent/docker/healthcheck.sh
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -48,3 +48,6 @@ VOLUME ["/etc/${COMPONENT}/jars", "/etc/${COMPONENT}/secrets"]
 COPY include/etc/confluent/docker /etc/confluent/docker
 
 CMD ["/etc/confluent/docker/run"]
+
+HEALTHCHECK --interval=5s --timeout=10s --retries=120 \
+	CMD /etc/ksql-server/healthcheck.sh 
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -50,4 +50,4 @@ COPY include/etc/confluent/docker /etc/confluent/docker
 CMD ["/etc/confluent/docker/run"]
 
 HEALTHCHECK --interval=5s --timeout=10s --retries=120 \
-	CMD /etc/ksql-server/healthcheck.sh 
+	CMD /etc/confluent/docker/healthcheck.sh
--- a/debian/kafka-connect-base/Dockerfile
+++ b/debian/kafka-connect-base/Dockerfile
@@ -49,5 +49,10 @@ COPY include/etc/confluent/docker /etc/confluent/docker
 
 CMD ["/etc/confluent/docker/run"]
 
-HEALTHCHECK --interval=5s --timeout=10s --retries=120 \
+# Polling period  : 5 seconds
+# Timeout period  :10 seconds (if the polling does not return within this time, treat as a failed poll)
+# Start-up period : 2 minutes (during which failures are not counted as failures)
+# Retry period    : 8 minutes (after which container is deemed unhealthy)
+# All settings can be overriden at run-time in Docker/Docker Compose. 
+HEALTHCHECK --start-period 120s --interval=5s --timeout=10s --retries=96 \
 	CMD /etc/confluent/docker/healthcheck.sh
--- a/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+if [[ -z $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM ]]; then
+  if [[ $(curl -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+    echo "Woohoo! Kafka Connect is up!"
+    exit 0
+  else 
+    echo -e $(date) "\tKafka Connect HTTP state: " $(curl -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/info) " (waiting for 200)"
+    exit 1
+  fi
+else
+  if [[ $(curl -k -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+    echo "Woohoo! Kafka Connect with SSL listener is up!"
+    exit 0
+  else 
+    echo -e $(date) "\tKafka Connect with SSL listener HTTP state: " $(curl -k -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/info) " (waiting for 200)"
+    exit 1
+  fi
+fi
--- a/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
@@ -1,19 +1,19 @@
 #!/usr/bin/env bash
 
 if [[ -z $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM ]]; then
-  if [[ $(curl -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+  if [[ $(curl -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/connectors) = 200 ]]; then
     echo "Woohoo! Kafka Connect is up!"
     exit 0
   else 
-    echo -e $(date) "\tKafka Connect HTTP state: " $(curl -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) " (waiting for 200)"
+    echo -e $(date) "\tKafka Connect HTTP state: " $(curl -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/connectors) " (waiting for 200)"
     exit 1
   fi
 else
-  if [[ $(curl -k -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+  if [[ $(curl -k -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/connectors) = 200 ]]; then
     echo "Woohoo! Kafka Connect with SSL listener is up!"
     exit 0
   else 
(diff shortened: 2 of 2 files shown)


debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
```diff
@@ -0,0 +1,19 @@
...
+  fi
+else
+  if [[ $(curl -k -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+    echo "Woohoo! Kafka Connect with SSL listener is up!"
+    exit 0
+  else 
+    echo -e $(date) "\tKafka Connect with SSL listener HTTP state: " $(curl -k -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/info) " (waiting for 200)"
+    exit 1
+  fi
+fi
```
wicknicks: BTW, should we check if a connector was running or not? the REST endpoint will continue to work once the worker comes up. If any start up step fails (status topics not available, for example), I think the worker should die immediately.

debian/kafka-connect-base/Dockerfile
```diff
@@ -48,3 +48,6 @@ VOLUME ["/etc/${COMPONENT}/jars", "/etc/${COMPONENT}/secrets"]
 COPY include/etc/confluent/docker /etc/confluent/docker
 
 CMD ["/etc/confluent/docker/run"]
+
+HEALTHCHECK --interval=5s --timeout=10s --retries=120 \
```
wicknicks: maybe we want to use start-period here to work with slower/smaller VMs? since the rest server comes up after  plugins are loaded?

debian/kafka-connect-base/Dockerfile
```diff
@@ -48,3 +48,6 @@ VOLUME ["/etc/${COMPONENT}/jars", "/etc/${COMPONENT}/secrets"]
...
+	CMD /etc/ksql-server/healthcheck.sh
```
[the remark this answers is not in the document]
rmoff: d'oh! Fixed.

debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+if [[ -z $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM ]]; then
+  if [[ $(curl -s -o /dev/null -w %{http_code} http://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/) = 200 ]]; then
+    echo "Woohoo! Kafka Connect is up!"
+    exit 0
+  else 
+    echo -e $(date) "\tKafka Connect HTTP state: " $(curl -s -o /dev/null -w %{http_code} $CONNECT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM://$CONNECT_REST_ADVERTISED_HOST_NAME:$CONNECT_REST_PORT/info) " (waiting for 200)"
```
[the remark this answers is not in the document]
rmoff: d'oh! Fixed.

debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
rmoff: Not sure I follow - do you mean check for an actual connector? As a base image there'll not be a connector created a lot of the time.

debian/kafka-connect-base/Dockerfile
rmoff: Could do, but reading the docs I think `retries` will achieve the same here. We'll poll for a total of ten minutes (5*120) and count as 'unhealthy' after that. WDYT?

wicknicks: I suppose that depends on the use case. A longer retry timeout means we'll report errors much later. Also, what errors do we plan to catch once the Rest API comes up though?

debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
wicknicks: In that case, yes, we won't check the connectors, but this healthcheck is inherited by the connect and replicator images so they would benefit from it (once of course connectors are instantiated)?

debian/kafka-connect-base/Dockerfile
rmoff: > Also, what errors do we plan to catch once the Rest API comes up though?

None at all. This healthcheck just ensures that the Kafka Connect REST API is up. It's not comprehensive, but it's 100% more than we currently have :)

debian/kafka-connect-base/include/etc/confluent/docker/healthcheck.sh
rmoff: Yes they would benefit from that. Sounds like a v2 of the healthcheck :) For now this is just basic functionality to check the REST API.

ybyzek:
Cool, it takes into account HTTPS as well has HTTP!  Nice work @rmoff 

Questions:

1. Is this meant to replace the cub's `connect-ready` python script?

2. Connectors cannot be submitted only when the REST endpoint is ready—the "Herder" also needs to be ready.  The bigger value add of this healthcheck over cub's `connect-ready` would be to check when the container is ready to accept connectors.  For your consideration: augment the healthcheck to be ready to run connectors.

rmoff:
Thanks @ybyzek 

> 1. Is this meant to replace the cub's `connect-ready` python script?

No it's not

> 2. Connectors cannot be submitted only when the REST endpoint is ready—the "Herder" also needs to be ready.  The bigger value add of this healthcheck over cub's `connect-ready` would be to check when the container is ready to accept connectors.  For your consideration: augment the healthcheck to be ready to run connectors.

Hmm interesting. Do you have any docs/testing that shows what happens if you try to use the rest endpoint when herder isn't ready? Sounds initially like a limitation with Kafka Connect if it's exposing its REST API before it's actually ready to handle requests.

ybyzek:
@rmoff 

1. Both check if the REST endpoint is up, so I'm not sure why it's not a replacement (for clarification: I mean a logical replacement -- a person wouldn't use both `connect-ready` and this `healthcheck`, right?)

2. This is why this wait is here: https://github.com/confluentinc/cp-docker-images/blob/5.3.0-post/examples/multi-datacenter/start.sh#L9-L16  Not sure why the REST endpoint comes up before it's ready to receive tasks—that's a question for the eng team.  But if the connector is submitted to early, either the REST endpoint isn't up (which this healthcheck would catch), or the task submission fails, e.g. repro'd below:

```bash
scripts/connectors/submit_wikipedia_irc_config.sh
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /connectors. Reason:
<pre>    Not Found</pre></p><hr><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.18.v20190429</a><hr/>

</body>
</html>
```

rmoff:
> 1. Both check if the REST endpoint is up, so I'm not sure why it's not a replacement (for clarification: I mean a logical replacement -- a person wouldn't use both `connect-ready` and this `healthcheck`, right?)

Ah I see what you mean. Yes in that case it is the same functionality. 

> 2. This is why this wait is here: https://github.com/confluentinc/cp-docker-images/blob/5.3.0-post/examples/multi-datacenter/start.sh#L9-L16  Not sure why the REST endpoint comes up before it's ready to receive tasks—that's a question for the eng team.  But if the connector is submitted to early, either the REST endpoint isn't up (which this healthcheck would catch), or the task submission fails, e.g. repro'd below:

Makes sense - so then instead of waiting for `200` from `/` we should wait for `200` from `/connectors` (since per your example it throws `404`)
@wicknicks any comments on this before I make the change?

ybyzek:
@rmoff 

> Makes sense - so then instead of waiting for 200 from / we should wait for 200 from /connectors (since per your example it throws 404)

My notes are based only on observation.  It's better to check with engineering what the precise REST endpoint is that determines readiness for users to start submitting connectors.

debian/kafka-connect-base/Dockerfile
rspurgeon: +1 for `--start-period`.  We know there is a startup delay so I think it makes sense to give the container time to boot without counting the first couple failures "against it" and it makes the actual failure count more accurate.

10 minutes to report unhealthy seems like a long time, although being able to overide in `docker run` and `docker-compose` maybe longer is better?

rmoff:
I've added `start-period`, and changed the test to be against `/connectors`. I've rebuilt and tested the image locally to verify the change. 

@rspurgeon @wicknicks could I get a +1 on this please so that I can go ahead and merge? :)

PR #772, opened 2019-07-26
Set correct KAFKA_VERSION value

robert-toth:
The KAFKA_VERSION seems to have been set to the confluent version and not the version of the Kafka broker used.


robert-toth:
[clabot:check]

PR #775, opened 2019-08-03
Fix plugin path

rmoff:
This PR sets the `CONNECT_PLUGIN_PATH` environment variable as part of the docker image for those in which connectors are installed (enterprise-replicator, server-connect, kafka-connect). Without it the user must remember to set it themselves (and be expected to know the value to which it should be set) otherwise their connectors will fail with Class errors. 

See Slack discussion here https://confluent.slack.com/archives/C5M8DKXJT/p1564768380084400


debian/kafka-connect/Dockerfile
```diff
@@ -39,3 +39,5 @@ RUN echo "===> Installing JDBC, Elasticsearch and Hadoop connectors ..." \
 
 RUN echo "===> Installing GCS Sink Connector ..."
 RUN confluent-hub install confluentinc/kafka-connect-gcs:latest --no-prompt
+
+ENV CONNECT_PLUGIN_PATH=/usr/share/java/,/usr/share/confluent-hub-components/
```
agam: nit: Please add a newline at the end of the file (here and for the one within `server-connect` too)

rmoff: Done

rmoff:
@agam I've squashed the commits, can you approve?

debian/enterprise-replicator/Dockerfile
```diff
@@ -30,3 +30,5 @@ RUN echo "===> Installing Replicator ..." \
         confluent-kafka-connect-replicator=${CONFLUENT_VERSION}${CONFLUENT_PLATFORM_LABEL}-${CONFLUENT_DEB_VERSION} \
      && echo "===> Cleaning up ..."  \
      && apt-get clean && rm -rf /tmp/* /var/lib/apt/lists/*
+
+ENV CONNECT_PLUGIN_PATH=/usr/share/java/kafka-connect-replicator
```
rhauch: This is not the correct plugin path, is it? See https://docs.confluent.io/current/connect/userguide.html#connect-installing-plugins

In this particular case, I'd expect the `/usr/share/java` directory is the plugin path, so that Connect considers each directory inside this (e.g., `/usr/share/java/kafka-connect-replicator`) as a _plugin directory in which it can find the JARs.

More specifically, I'd expect the following directory structure in this image
```
/
+- usr
     +- share
          +- java     <--- This is the parent directory of the plugin directories and therefore should be on the plugin path
            +- kafka-connect-replicator     <--- This is a plugin directory
                 +- Replicator JAR files
            +- kafka-connect-jdbc (if it were installed)      <--- This would be a plugin directory
            +- kafka-connect-elasticsearch (if it were installed)     <--- This would a plugin directory
            +- etc
```
In fact, when I start a container with the latest image, this is what I see:
```
$ docker run -it confluentinc/cp-enterprise-replicator:latest /bin/sh
# cd /usr/share/java
# ls -la
total 52
drwxr-xr-x 1 root root 4096 Feb 18 23:50 .
drwxr-xr-x 1 root root 4096 Feb 18 23:36 ..
drwxr-xr-x 2 root root 4096 Feb 18 23:33 confluent-common
drwxr-xr-x 2 root root 4096 Feb 18 23:36 confluent-control-center
drwxr-xr-x 2 root root 4096 Feb 18 23:36 confluent-hub-client
drwxr-xr-x 2 root root 4096 Feb 18 23:36 confluent-rebalancer
-rw-r--r-- 1 root root  957 May  6  2014 java_defaults.mk
drwxr-xr-x 1 root root 4096 Feb 18 23:36 kafka
drwxr-xr-x 2 root root 4096 Feb 18 23:50 kafka-connect-replicator
drwxr-xr-x 2 root root 4096 Feb 18 23:33 kafka-serde-tools
drwxr-xr-x 2 root root 4096 Feb 18 23:36 monitoring-interceptors
drwxr-xr-x 2 root root 4096 Feb 18 23:33 rest-utils
drwxr-xr-x 2 root root 4096 Feb 18 23:33 schema-registry
# cd kafka-connect-replicator
# ls -la
total 28052
drwxr-xr-x 2 root root    4096 Feb 18 23:50 .
drwxr-xr-x 1 root root    4096 Feb 18 23:50 ..
-rw-r--r-- 1 root root   85912 Feb 18 11:36 airline-0.7.jar
-rw-r--r-- 1 root root   77185 Feb 18 11:36 annotations-2.0.3.jar
-rw-r--r-- 1 root root   20437 Feb 18 11:36 audience-annotations-0.5.0.jar
-rw-r--r-- 1 root root   20905 Feb 18 11:36 common-utils-5.1.2.jar
-rw-r--r-- 1 root root  284184 Feb 18 11:36 commons-codec-1.10.jar
-rw-r--r-- 1 root root   37932 Feb 18 11:36 confluent-licensing-new-5.1.2.jar
-rw-r--r-- 1 root root  811686 Feb 18 11:36 confluent-serializers-new-5.1.2.jar
-rw-r--r-- 1 root root  231952 Feb 18 11:36 gson-2.7.jar
-rw-r--r-- 1 root root 2442625 Feb 18 11:36 guava-20.0.jar
-rw-r--r-- 1 root root   66519 Feb 18 11:36 jackson-annotations-2.9.0.jar
-rw-r--r-- 1 root root  325619 Feb 18 11:36 jackson-core-2.9.8.jar
-rw-r--r-- 1 root root 1347236 Feb 18 11:36 jackson-databind-2.9.8.jar
-rw-r--r-- 1 root root    2497 Feb 18 11:36 javax.inject-1.jar
-rw-r--r-- 1 root root   87325 Feb 18 11:36 jline-0.9.94.jar
-rw-r--r-- 1 root root   78146 Feb 18 11:36 jopt-simple-5.0.4.jar
-rw-r--r-- 1 root root  259680 Feb 18 11:36 jose4j-0.6.4.jar
-rw-r--r-- 1 root root  139472 Feb 18 11:36 kafka-connect-replicator-5.1.2.jar
-rw-r--r-- 1 root root 7571742 Feb 18 11:36 kafka_2.11-2.1.1-cp1.jar
-rw-r--r-- 1 root root  489884 Feb 18 11:36 log4j-1.2.17.jar
-rw-r--r-- 1 root root   82123 Feb 18 11:36 metrics-core-2.2.0.jar
-rw-r--r-- 1 root root 1292696 Feb 18 11:36 netty-3.10.6.Final.jar
-rw-r--r-- 1 root root 1383882 Feb 18 11:36 protobuf-java-3.4.0.jar
-rw-r--r-- 1 root root   68545 Feb 18 11:36 protobuf-java-util-3.4.0.jar
-rw-r--r-- 1 root root 5749423 Feb 18 11:36 scala-library-2.11.12.jar
-rw-r--r-- 1 root root   82031 Feb 18 11:36 scala-logging_2.11-3.9.0.jar
-rw-r--r-- 1 root root 4623075 Feb 18 11:36 scala-reflect-2.11.12.jar
-rw-r--r-- 1 root root    8869 Feb 18 11:36 slf4j-log4j12-1.7.5.jar
-rw-r--r-- 1 root root   74798 Feb 18 11:36 zkclient-0.10.jar
-rw-r--r-- 1 root root  906708 Feb 18 11:36 zookeeper-3.4.13.jar
```

So the plugin path should be set to `/usr/share/java`. However, the following files all have `plugin.path=/usr/share/java` already set:
```
/etc/schema-registry/connect-avro-distributed.properties
/etc/kafka/connect-distributed.properties
```
but the following file has this commented out:
```
/etc/kafka-connect-replicator/replicator-connect-distributed.properties
```
That means that if we're starting replicator using the `/etc/kafka-connect-replicator/replicator-connect-distributed.properties` file for the Connect distributed worker, then we do need to set the plugin path. However, if we're using one of the other Connect distributed configuration files, we shouldn't have to, as the plugin path should already be set up.

rhauch:
Okay, so I traced more about how this is being set up. It does look like the https://github.com/confluentinc/cp-docker-images/blob/5.3.0-post/examples/multi-datacenter/docker-compose.yml sets `CONNECT_PLUGIN_PATH: /usr/share/java`, and I cannot see where the Replicator executable run via the Docker image is properly setting the plugin path.

So, we need to set `CONNECT_PLUGIN_PATH: /usr/share/java` in these places.

debian/enterprise-replicator/Dockerfile
rhauch: This is what I think it should be:

```suggestion
ENV CONNECT_PLUGIN_PATH=/usr/share/java/
```

debian/kafka-connect/Dockerfile
```diff
@@ -39,3 +39,6 @@ RUN echo "===> Installing JDBC, Elasticsearch and Hadoop connectors ..." \
...
```
rhauch: This is good: the `/usr/share/java/` is correct, and the `/usr/share/confluent-hub-components` is where the Confluent Hub client will install additional connectors.

rhauch: Actually, should this be set in `debian/kafka-connect-base` instead of here? That way images based on that image (`cp-kafka-connect` and customer-specific images) will have this property set correctly.

debian/server-connect/Dockerfile
```diff
@@ -39,3 +39,6 @@ RUN echo "===> Installing JDBC, Elasticsearch and Hadoop connectors ..." \
 
 RUN echo "===> Installing GCS Sink Connector ..."
 RUN confluent-hub install confluentinc/kafka-connect-gcs:latest --no-prompt
+
+ENV CONNECT_PLUGIN_PATH=/usr/share/java/,/usr/share/confluent-hub-components/
```
rhauch: Actually, should this be set in `debian/server-connect-base` instead of here? That way images based on that image (e.g., `cp-enterprise-replicator` and `cp-enterprise-replicator-executable`) will have this property set correctly.

debian/enterprise-replicator/Dockerfile
rmoff: Done

ewencp:
> Since both the commits inside this PR have the same description, please either (1) squash them, or (2) add a bit more to each to distinguish them (`Add plugin path ENV for ...`)

For reference, you generally don't need to do this manually anymore if you don't want to. Use the `Squash and merge` option when merging and GH does it for you. Usually shouldn't be an issue since PR should be small and self contained.

debian/server-connect/Dockerfile
rmoff: Done.

debian/kafka-connect/Dockerfile
rmoff: Done

rmoff:
@rhauch am I ok to merge this change? thanks.

PR #782, opened 2019-08-23
Set the SCALA_VERSION to 2.12.

yangxi:


mohnishbasha (commented):
lgtm, however could you get an ok from kafka team as well?

ewencp (approved):
If it is fine in the regular dockerfile, should be fine here, though tbh I'm surprised we're changing stuff here since these aren't really supported?

ijuma (approved):
LGTM

PR #794, opened 2019-09-16
add security plugins for schema validator to cp-server

sdanduConf:
With Schema Validator, we now need the security-plugins installed on the broker image itself.  This will install confluent-security into the cp-server image.

Signed-off-by: sdanduConf <sdandu@confluent.io>


ewencp:
Keep in mind that there are PRs in flight to move these dockerfiles into separate repos, so you'll need to make sure you keep in sync with those to avoid losing this change.

sdanduConf:
> Keep in mind that there are PRs in flight to move these dockerfiles into separate repos, so you'll need to make sure you keep in sync with those to avoid losing this change.

Thanks, will create a JIRA blocker for 5.4 to make sure that doesn't get lost and sync with DevProd to make sure i understand the changes.

sdanduConf:
took some time, but did some manual validation to make sure that the image had confluent-security-plugins installed and basic schema validation setting and validation worked.

PR #802, opened 2019-10-01
adding a zookeeper observer property to zookeeper.properties.template

addisonhuddy:
Adding a property to allow zookeeper to run as an observer.


joel-hamill:
@addisonhuddy or ping the [Slack #docker channel](https://confluent.slack.com/messages/C1U1RGWA1)

gwenshap:
Guessing it probably wasn't that important?

PR #805, opened 2019-10-07
Fix for issue 797

### LINKED ISSUE #797: Connect's Internal converters configurations still required
It appears that https://github.com/confluentinc/cp-docker-images/pull/761/ is incomplete.  It removed the check that those variables are defined (`dub ensure`) with the intention that docker images could remove the configs:

```
CONNECT_INTERNAL_KEY_CONVERTER: "org.apache.kafka.connect.json.JsonConverter"
CONNECT_INTERNAL_VALUE_CONVERTER: "org.apache.kafka.connect.json.JsonConverter"
```

However removing those config lines results in the Connect image failing:

```
kafka-connect      | /etc/confluent/docker/configure: line 36: CONNECT_INTERNAL_KEY_CONVERTER: unbound variable
```

The reason is that the code in the kafka-connect-base image references those variables anyway, and thus will fail if they're not set: https://github.com/confluentinc/cp-docker-images/blob/5.3.1-post/debian/kafka-connect-base/include/etc/confluent/docker/configure#L36. 

cc: @rmoff

rmoff:
Fix #797. Now pass in an empty default so that if the variable isn't defined (which it doesn't have to be any more) the script doesn't fail.

### DIFF
--- a/debian/kafka-connect-base/include/etc/confluent/docker/configure
+++ b/debian/kafka-connect-base/include/etc/confluent/docker/configure
@@ -36,7 +36,7 @@ if [ -z "$CONNECT_REST_PORT" ]; then
 fi
 
 # Fix for https://issues.apache.org/jira/browse/KAFKA-3988
-if [[ $CONNECT_INTERNAL_KEY_CONVERTER == "org.apache.kafka.connect.json.JsonConverter" ]] || [[ $CONNECT_INTERNAL_VALUE_CONVERTER == "org.apache.kafka.connect.json.JsonConverter" ]]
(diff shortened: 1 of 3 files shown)


rmoff:
It sounds like this is going to need changing/fixing again on `kafka-connect-images` -- @confluentinc/connect can perhaps advise how&where to make this change as part of the migration?

rmoff:
Thanks @wicknicks ! Can I just merge this into 5.0.x and `pint` it through? or does it need merging into `kafka-connect-images` instead (or as well)?

elismaga:
> Thanks @wicknicks ! Can I just merge this into 5.0.x and `pint` it through? or does it need merging into `kafka-connect-images` instead (or as well)?

@rmoff @wicknicks If you want this change to go into 5.0.x you can commit it there and then pint merge all the way to 5.3.x. To get this change into the the master branch / 5.4.0 you will need to make a separate PR to commit this change to the kafka-connect-images repo. Sorry for the extra work. If you don't need to back port this to 5.0.x then you can just push this to kafka-connect-images repo only and close this PR.

PR #823, opened 2020-01-09
print() is a function in modern Python

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

__xrange()__ was removed in Python 3 in favour of an iterator-base version of __range()__.  In Python 2, for low values for __n__ there is no appreciable difference __xrange(n)__ and __range(n)__.

### DIFF
--- a/tests/test_enterprise_kafka.py
+++ b/tests/test_enterprise_kafka.py
@@ -111,7 +111,7 @@ class ClusterHostNetworkTest(unittest.TestCase):
         rebalance_status = self.cluster.run_command_on_service("kafka-1", ADB_STATUS)
 
         rebalance_complete = ""
-        for i in xrange(120):
+        for i in range(120):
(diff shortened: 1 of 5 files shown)


cricket007:
Are these the only issues?

You could try running 2to3 on the script

cclauss:
2to3 is bad idea.  It often breaks Python 2 functionality to achieve Python 3 compatibility.  Use [futuzize](https://pypi.org/project/future) or [modernize](https://pypi.org/project/modernize) instead if you want to retain compatibility with Python 2.

cclauss:
This PR is the syntax errors plus the undefined names.  I would recommend landing these change and testing functionality before adding other changes.

cricket007:
Fair enough, I was just suggesting upgrading the whole script to py3, not exactly prescribing the solution

PR #847, opened 2020-04-17
fix: avoid leak of aws credentials on log

robsonpeixoto:
show_env is showing all AWS credentials on the log and it is a security problem.


elijah-roberts:
@robsonpeixoto Do you want to sigh the CLA and see if we can git this merged in? We are seeing this issue as well...

PR #868, opened 2020-07-28
Added --force-yes in Dockefile

ankurgarg53:
Added --force-yes to get rid of the following Error while building (E: There are problems and -y was used without --force-yes)
OS- centos-release-7-6.1810.2.el7.centos.x86_64


ankurgarg53:
> [clabot:check]

